PR #2, opened 2026-02-25
fix: node-pty spawn-helper permissions on macOS with pnpm

r-dent:
pnpm doesn't preserve execute bits from node-pty prebuilds, causing posix_spawnp to fail when spawning PTY processes.

- Add postinstall script to chmod +x spawn-helper for darwin-*
- Skip otool Metal detection when whisper-cli isn't installed


puritysb (approved):
LGTM! Clean fixes for real-world install issues:

- **spawn-helper chmod**: pnpm strips execute bits from node-pty prebuilds — postinstall `chmod +x` is the right fix for macOS
- **detectMetal guard**: `existsSync()` early return prevents crash when whisper-cli isn't installed

Thanks for the contribution @r-dent!

Issue #3, opened 2026-03-19
posix_spawnp failed on Node.js 24 — prebuilt node-pty binary incompatible


eprouveze:
## Problem

Running `sdc` on Node.js v24.2.0 (macOS 15, Apple Silicon) fails immediately:

```
[sdc] Starting AgentDeck bridge on port 9120 (agent: claude-code)...
[sdc] Failed to start adapter: Error: posix_spawnp failed.
```

## Root Cause

The prebuilt `node-pty` binary (`prebuilds/darwin-arm64/pty.node` and `spawn-helper`) shipped with `@agentdeck/bridge` is incompatible with Node.js 24. The native addon fails at the `posix_spawnp` syscall when trying to spawn a PTY.

## Fix

Rebuilding `node-pty` from source resolves the issue:

```bash
cd $(npm root -g)/@agentdeck/bridge/node_modules/node-pty
npx node-gyp rebuild
```

After rebuild, `sdc` starts normally and works perfectly (tested with Warp terminal, not just iTerm2).

## Suggestion

Consider adding a postinstall check in `@agentdeck/setup` that detects Node 24+ and triggers a source rebuild of `node-pty` if the prebuilt binary fails a quick spawn test. Something like:

```js
try {
  const pty = require('node-pty');
  const p = pty.spawn('/bin/echo', ['test'], { cols: 80, rows: 24 });
  p.kill();
} catch {
  console.log('[setup] Prebuilt node-pty failed, rebuilding from source...');
  execSync('npx node-gyp rebuild', { cwd: 'node_modules/node-pty' });
}
```

## Environment

- macOS 15.4 (Sequoia)
- Node.js v24.2.0
- Apple M4 (Mac Mini)
- Terminal: Warp (works fine after rebuild — iTerm2 not required)
- AgentDeck installed via `npx @agentdeck/setup`

puritysb:
Thanks for the detailed report and the clear workaround! 🙏

We've addressed this in a few ways:

### Changes

1. **Node.js >= 22 required** — Node 20 hits EOL in April 2026, so we've bumped the minimum. Node 22 and 24 (both LTS) are now the supported versions.

2. **Source build by default** — `npx @agentdeck/setup` now sets `npm_config_build_from_source=true` when installing `@agentdeck/bridge`, so node-pty is always compiled from source against your current Node.js version. This eliminates prebuilt ABI mismatch entirely.

[...]

PR #4, opened 2026-04-10
fix: session slots empty on first launch (4 root causes)

pakuda:
## Summary

Session slot buttons were always showing \"Empty\" after installing the plugin and running `agentdeck claude`. Diagnosed and fixed 4 root causes:

- **Missing `shared/src/svg-renderers/`** — commit `8db82ff` added re-exports from this folder but never committed the actual files, breaking `pnpm build` with TS2307. Reconstructed the 4 missing files from the previous commit (`8cdaef6`).
- **Plugin didn't re-send `deck_slot_map` on reconnect** — `sendSlotMap()` is triggered by `onWillAppear` (~5 s after plugin load). When the bridge starts *after* the plugin the first send is dropped. The `connected` handler only sent `query_usage` and never re-sent the slot map, so the bridge had no record of the 8 button slots.
- **Bridge didn't re-broadcast `sessions_list` after receiving slot map** — even with a valid slot map, the bridge only broadcast `button_state` on `deck_slot_map`, not `sessions_list`. Session-slot buttons populate from `sessions_list`, so they stayed empty.
- **`buildEnrichedSessionsList` excluded the own session** — the aggregator filtered `s.id !== ownSessionId`, producing an empty list in single-session mode (`agentdeck claude` without a daemon). Fix: include all non-daemon sessions so single-session mode works. The daemon-server enricher is unaffected.

## Test plan

- [ ] `pnpm build` succeeds from a clean checkout
- [ ] Run `agentdeck claude` in a project directory
- [ ] Session slot button 0 shows the project name + state within a few seconds of connection
- [ ] Restart Stream Deck app while bridge is running — slots repopulate after reconnect
- [ ] Daemon multi-session mode (`agentdeck daemon start` + multiple `agentdeck claude` sessions) still shows all sessions correctly

### DIFF
--- a/bridge/src/index.ts
+++ b/bridge/src/index.ts
@@ -671,6 +671,8 @@ export async function startSession(opts: SessionOptions): Promise<void> {
       core.wsServer.broadcastExcept(cachedSlotMap, sender);
       broadcastSse(cachedSlotMap);
       core.broadcast(computeButtonState());
+      // Re-send sessions_list so slot buttons populate after plugin startup race
+      core.broadcastSessionsList().catch(() => {});
       return true;
     }
     return false;
--- a/bridge/src/session-aggregator.ts
+++ b/bridge/src/session-aggregator.ts
@@ -44,12 +44,13 @@ export async function enrichSessionsWithState(
 
 /**
  * Build an enriched sessions list for multi-session display.
+ * Always includes the own session so single-session mode (agentdeck claude) shows the active session.
  */
 export async function buildEnrichedSessionsList(
   ownSessionId: string,
   ownState: string,
 ): Promise<EnrichedSession[]> {
-  const siblings = listActiveSessions().filter(s => s.agentType !== 'daemon' && s.id !== ownSessionId);
-  const enriched = await enrichSessionsWithState(siblings, ownSessionId, ownState);
+  const all = listActiveSessions().filter(s => s.agentType !== 'daemon');
+  const enriched = await enrichSessionsWithState(all, ownSessionId, ownState);
   return sortSessions(enriched);
 }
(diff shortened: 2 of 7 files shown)


puritysb (changes requested):
Thanks for the detailed root-cause analysis — all four causes are real and the fixes look logically correct. The SVG renderer additions are clean (proper `escXml()` escaping on all user-controlled strings, no new dependencies).

Two things need to be resolved before this can merge:

### 1. Not currently mergeable — rebase required

The PR is based on `8db82ff`, but `master` has since advanced. In particular, commit `f7d7f4b2` modified `session-aggregator.ts` to add a third `ownModelName?: string` parameter to `buildEnrichedSessionsList`. You'll need to rebase onto the current `master` and resolve any conflicts there.

### 2. Existing test will fail after the `session-aggregator.ts` change

`bridge/src/__tests__/session-aggregator.test.ts` has a test named:

> `"buildEnrichedSessionsList excludes daemon and own session before enrichment"`

It asserts `toHaveLength(2)` (siblings only). After your fix, the own session is included, so it will now return 3. Please update the test name and assertion to reflect the new intended behavior (include own session, exclude daemon). A new test case covering the single-session scenario would also be welcome.

Once those two are addressed this looks good to merge.

puritysb:
Thanks for the careful root-cause analysis here, @pakuda — the fixes for #2, #3, and #4 were spot-on and the bugs were still live on `master`. I've rebased the change onto current `master` and reduced the scope (the `shared/src/svg-renderers/*` reconstruction is already covered by `0d8da018` and follow-ups), and updated the `session-aggregator.test.ts` case as the review asked. Your diagnosis credit is preserved via `Co-Authored-By`.

Continuing in #5 — closing this one in favor of that.

puritysb:
Superseded by #5.

PR #5, opened 2026-04-28
fix(slots): session slots empty on first launch — startup race + own-session filter

puritysb:
## Summary

Re-takes #4 — rebased onto current `master` and reduced to the three still-applicable fixes. Diagnosis credit to @pakuda; preserved via `Co-Authored-By` trailer.

Three independent root causes left session-slot buttons stuck on "Empty" after a clean install / first launch:

- **Plugin didn't re-send `deck_slot_map` on reconnect** — `sendSlotMap()` is triggered by `onWillAppear` (~5 s after plugin load). When the bridge starts *after* the plugin the first send is dropped. The `connected` handler only sent `client_register` + `query_usage` and never re-sent the slot map, so the bridge had no record of the button slots. Fix: call `sendSlotMap()` at the top of the `connected` handler.
- **Bridge didn't re-broadcast `sessions_list` after `deck_slot_map`** — even with a valid slot map, the bridge only broadcast `button_state`. Session-slot buttons populate from `sessions_list`, so they stayed empty. Fix: call `core.broadcastSessionsList()` in the `deck_slot_map` raw message handler.
- **`buildEnrichedSessionsList` excluded the own session** — the aggregator filtered `s.id !== ownSessionId`, producing an empty list in single-session mode (`agentdeck claude` without a daemon). Fix: include all non-daemon sessions; the daemon-server enricher continues to override this list in multi-session setups.

The `shared/src/svg-renderers/*` reconstruction from #4 is dropped — `0d8da018` already added the missing files and they've evolved further in `08ec688c` / `b7e682ae` / `38dd7f46` / `56ab762d` / `a50c9b23`.

## Test plan

- [x] `pnpm -w typecheck` — clean
- [x] `pnpm -w build` — clean
- [x] `pnpm test` — 1015/1015 passing (incl. updated + new aggregator cases)
- [ ] `agentdeck claude` standalone — slot 0 shows project name + state within ~5 s of connection
- [ ] Restart Stream Deck app while bridge is running — slots repopulate after reconnect
- [ ] `agentdeck daemon start` + multiple `agentdeck claude` sessions — all sessions appear

`session-aggregator.test.ts`:
- existing case `"buildEnrichedSessionsList excludes daemon and own session before enrichment"` renamed to `"includes own session and excludes daemon"`, length `2 → 3`, ordering reflects `sortSessions` (`Backend` < `Main` within `claude-code` group, then `codex-cli`)
- new case `"returns own session in single-session mode without /health calls"` covers the standalone `agentdeck claude` path that #4 was originally diagnosing

Closes #4.

### DIFF
--- a/bridge/src/__tests__/session-aggregator.test.ts
+++ b/bridge/src/__tests__/session-aggregator.test.ts
@@ -158,7 +158,7 @@ describe('session-aggregator', () => {
     expect(sessions[0].state).toBeUndefined();
   });
 
-  it('buildEnrichedSessionsList excludes daemon and own session before enrichment', async () => {
+  it('buildEnrichedSessionsList includes own session and excludes daemon', async () => {
     mockListActive.mockReturnValue([
       makeSession({ id: 'own-session', port: 9121, projectName: 'Main' }),
       makeSession({ id: 'daemon-1', port: 9120, projectName: 'Daemon', agentType: 'daemon' }),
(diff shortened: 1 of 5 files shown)


puritysb:
Codex stop-time review caught that the prior commit only wired the rebroadcast in the per-session bridge. Since the plugin connects daemon-only (see \`plugin/src/connection-manager.ts\`), the multi-session daemon path was still subject to the race — slot buttons would only populate after the 10 s sessions polling tick. Pushed \`98080a17\` to mirror the same handler in \`daemon-server.ts\`'s \`onRawMessage\`.

\`pnpm test\` 1015/1015 still green, typecheck + build clean.

PR #22, opened 2026-06-26
fix(mdns): pin multicast egress to LAN interface on Windows

doug-w:
## Problem

On a Windows Hyper-V/WSL host, the bridge has several IPv4 interfaces — the real LAN adapter plus host-only virtual switches (172.x) and APIPA (169.254.x). `multicast-dns` (under `bonjour-service`) bound the mDNS socket to `0.0.0.0` and let Windows pick the outbound multicast interface, which was a **Hyper-V virtual switch**. The service announcement therefore never egressed on the real WiFi/LAN adapter, so **iOS clients on the LAN never discovered the bridge**.

## Fix

- `bridge/src/mdns.ts`: pass `interface: <default-route LAN ip>` to the `Bonjour` constructor on `win32`, so `multicast-dns` binds the socket, joins the group, and `setMulticastInterface` all on the physical LAN adapter. Gated to `win32` so macOS/Linux multi-interface discovery is unaffected.
- `shared/src/net-utils.ts`: resolve the default-route IPv4 correctly on Windows via `route print -4 0.0.0.0` (lowest-metric source IP), so the advertised TXT `ip=` and pairing URL point at the LAN adapter instead of a virtual switch.

## Verification (on the Hyper-V host)

`dns-sd -B _agentdeck._tcp local.`, before → after:

| | Interface the service multicasts on |
|---|---|
| Before | `if 64` only → `172.25.112.1` (Hyper-V switch) ❌ |
| After | `if 8` → `192.168.1.236` (real LAN) ✅ |

After the change the daemon's mDNS socket binds to `192.168.1.236:5353` (not `0.0.0.0`), the service resolves with TXT `ip=192.168.1.236`, and the iOS app now discovers it. Build clean; integration + net-utils tests green.

### DIFF
--- a/bridge/src/mdns.ts
+++ b/bridge/src/mdns.ts
@@ -95,9 +95,26 @@ export function advertiseBridge(
         instance = null;
       }
 
-      instance = new Bonjour();
-
       const lanIp = getLanIp();
+
+      // Windows multi-homed egress fix: a Hyper-V/WSL host has several IPv4
+      // interfaces (the real LAN adapter plus host-only virtual switches on
+      // 172.x and APIPA 169.254.x). Left to its own devices, `multicast-dns`
+      // joins the mDNS group on *all* of them and lets the OS pick the outbound
+      // multicast interface — which can be a virtual switch, so the announcement
+      // never egresses on the WiFi/LAN adapter and a remote iOS device never
+      // sees the service. Pinning `interface` to the default-route LAN IP makes
+      // multicast-dns bind the socket, addMembership, and setMulticastInterface
+      // all to that one adapter. Gated to win32 so macOS/Linux multi-interface
+      // discovery (where bonjour-service correctly fans out) is unaffected.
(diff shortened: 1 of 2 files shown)


puritysb (approved):
**Approve (minor).** Sound fix for Windows multi-homed mDNS egress.

- `bridge/src/mdns.ts`: on `win32` only, `new Bonjour({ interface: lanIp })` pins the socket bind / membership / `setMulticastInterface` to the physical LAN adapter — macOS/Linux multi-interface fan-out is preserved by the gating.
- `shared/src/net-utils.ts`: parsing `route print -4 0.0.0.0` is locale-independent (numeric pattern match), excludes `169.254.*`, and picks the lowest metric so VPN/host-only adapters lose. `directIp` is set only on `win32`, so the mac/Linux paths are unchanged.

Non-blocking: a unit test for the win32 `defaultRouteIp` parser (mocking `execFileSync`) would guard against `route print` output-format regressions.

PR #23, opened 2026-06-26
feat(bridge): Windows daemon autostart via per-user Scheduled Task

doug-w:
## Summary

Wires `agentdeck daemon install/uninstall` on Windows to register a per-user **Scheduled Task** (`AgentDeckDaemon`, logon trigger) via built-in `schtasks.exe` — the Windows analog of the macOS LaunchAgent. It runs in the interactive session (so USB-HID / audio / mDNS / Stream Deck app all work), needs no admin elevation, and adds no npm dependency. A session-0 Windows Service is intentionally **not** used.

## Changes

- **`bridge/src/windows-service.ts`**: pure `buildScheduledTaskXml()` (UTF-16+BOM, `RestartOnFailure`≈KeepAlive, `IgnoreNew`, `InteractiveToken`/`LeastPrivilege`, `node.exe` + quoted `cli.js` + `--foreground`) plus `schtasks` wrappers.
- **`bridge/src/cli.ts`**: win32 branches for install (register → run → Codex hooks, with fallback message) and uninstall (graceful `/shutdown` → `/Delete`, idempotent).
- **`bridge/src/daemon-server.ts`**: close a startup race exposed by autostart — when two daemon starts land within ~1s (concurrent logon-trigger fires), both passed the pre-bind singleton guard, then the loser hit `EADDRINUSE` and fell back to a fresh port, leaving two daemons and a clobbered `daemon.json`. Now on `EADDRINUSE` for the default port we re-probe `/health`: if a daemon already owns it, `exit(0)`; only fall back to a new port for a non-daemon occupant (session bridge), preserving prior behavior.
- **`setup/src/setup.ts`**: platform-neutral "optional daemon install" next-steps nudge.
- **`bridge/src/__tests__/windows-service.test.ts`**: unit tests for the XML builder.
- **docs**: `daemon.md` Autostart section, README + CLAUDE.md updates.

## Test

`windows-service.test.ts` covers the XML builder. Install/uninstall verified on Windows 11 (Scheduled Task registers, runs at logon, idempotent uninstall).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -119,7 +119,7 @@ This diagnostic path is developer tooling only: it lives in `scripts/` and `.age
 
 ## Windows dev setup
 
-The Node.js bridge, hook installer, and Stream Deck plugin run on Windows 11 (Apple/Android/ESP32 native builds are out of scope). Full prereqs, install/run steps, and the intentional Windows differences (ConPTY, data dir, PowerShell hook one-liner, daemon-install no-op, device-module gating, darwin-only sampler) live in **[README → Windows (Bridge + Plugin)](README.md#windows-bridge--plugin)**. Code refs: `bridge/src/pty-manager.ts`, `hooks/src/install.ts`, `bridge/src/cli.ts`.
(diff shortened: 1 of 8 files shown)


puritysb (approved):
**Approve (minor).** The design is right — a logon-trigger per-user Scheduled Task instead of a session-0 Windows Service (keeps USB-HID / audio / mDNS / the Stream Deck app working, no elevation, no npm dep) is the correct macOS-LaunchAgent analog.

- Nice to see `buildScheduledTaskXml` split out as a pure function with unit tests covering XML escaping / UTF-16 BOM / KeepAlive semantics.
- **The EADDRINUSE race fix in `daemon-server.ts` is a particularly good catch**: on concurrent start it re-probes the occupant's `/health` and `exit(0)`s only if `mode === 'daemon'`, falling back to a new port only for a non-daemon (session bridge) — preventing a clobbered `daemon.json`. `probeDaemonHealth` returns `{mode?, pid?}` (`session-registry.ts:297`), so it lines up. (Followed up in #32 to also verify the reported PID is a live, distinct process before conceding, so a forged/stale `mode:'daemon'` response can't block startup.)

Non-blocking:
- `WorkingDirectory = ~/.agentdeck` may not exist at first install → the task could fail to start; consider ensuring the directory (or confirming a missing cwd is tolerated).
- `process.execPath` pins the node path to whatever ran the install (acceptable — just confirming intent).

PR #33, opened 2026-06-27
fix(devices): re-apply dropped iDotMatrix Node module registration

puritysb:
## Why

A branch cleanup (the "Salvage … #30" commit) kept the **Apple + Phase-2 + cli** parts of the iDotMatrix/Timebox topology-visibility fix but **dropped the Phase-1 Node wiring**. Net effect on the **Node daemon**: iDotMatrix still rendered to the panel (legacy direct sync call) but became **invisible again** in `/health`, `/devices`, the TUI dashboard, and `agentdeck devices` — the exact bug the original fix addressed. (The Apple/Swift surface was unaffected.)

This PR surgically re-applies only the missing Node pieces.

## What

- **New `bridge/src/modules/idotmatrix-module.ts`** — mirrors `TimeboxModule` (`shouldActivate` / `start` owns sync + zero-config auto-discover / `stop` / `statusSnapshot`).
- **`modules/index.ts`** — register `IDotMatrixModule` in `createDefaultModules` (+ export).
- **`modules/types.ts`** — add `idotmatrix?: 'auto' | boolean` to `ModuleConfigs`.
- **`daemon-server.ts`** — add the `idotmatrix` branch to `buildNodeModuleHealth`, the `/devices` entry, and `idotmatrix: 'auto'` to the default module config; **remove** the now-redundant direct `startIDotMatrixSync` / `autoDiscoverIDotMatrix` / `stopIDotMatrixSync` calls so the module is the single owner (no double-spawn).
- **`DEVELOPMENT_LOG.md`** — restore the dropped entry documenting the 3-layer gap.

Timebox was already a proper module and is unaffected; this brings iDotMatrix to parity.

## Verification

- `pnpm build` green (bridge `tsc` clean); affected unit tests 22/22 pass (`server-integration`, `timebox-ble`).
- Node daemon e2e (temp HOME with a configured device): `/health.modules` now includes `idotmatrix` with device details and `/devices` includes the `idotmatrix` type.

### DIFF
--- a/DEVELOPMENT_LOG.md
+++ b/DEVELOPMENT_LOG.md
@@ -6,6 +6,26 @@
 
 ---
 
+## 2026-06-27 — iDotMatrix / Divoom Timebox Mini 가 대시보드·패널에 안 보이던 문제
+
+### 문제
+두 BLE 매트릭스 패널(iDotMatrix 32×32, Timebox Mini 11×11)이 대시보드 토폴로지에 안 뜨고 패널 렌더도 깨졌다. 단일 버그가 아니라 3겹 plumbing gap이었다.
+
+### 해결
+- **Node 가시성**: Timebox는 정식 `TimeboxModule`인데 **iDotMatrix는 모듈 미등록**(직접 `startIDotMatrixSync` spawn만)이라 `buildNodeModuleHealth`/`/devices`/TUI/`agentdeck devices` 전부에서 누락. `IDotMatrixModule` 신설(Timebox 미러)+`createDefaultModules` 등록+health/devices/cli/TUI 분기+`ModuleConfigs.idotmatrix`. daemon-server의 직접 start 호출은 제거(모듈이 단일 소유).
+- **Apple 가시성**: Swift 데몬은 `/health`에 idotmatrix+timebox `statusSnapshot()`을 이미 emit하지만 **클라가 버렸음** — `ModuleHealthState` 필드 없음+`BridgeEventParser.parseModuleHealth` 미디코드+`TopologyRail` 섹션 없음. 공유 `BLEMatrixHealth` struct+파서 분기+`bleMatrixSection`(statusReason→LEDStatus).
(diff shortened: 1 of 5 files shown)


puritysb (commented):
**The iDotMatrix module work LGTM — but please rebase onto `master` before merging (the branch doesn't build standalone).**

### The module itself (LGTM)
- `IDotMatrixModule` faithfully mirrors `TimeboxModule` (`shouldActivate` / `start` owns sync + gated zero-config discovery / `stop` / `statusSnapshot`), registered in `createDefaultModules`, with `buildNodeModuleHealth` + `/devices` + `ModuleConfigs.idotmatrix` wiring. Removing the direct `startIDotMatrixSync`/`autoDiscoverIDotMatrix`/`stopIDotMatrixSync` calls so the module is the single owner is the right call (no double-spawn).
- Minor: `start()` now gates auto-discovery on `isIDotMatrixAutoDiscoverEnabled()`, whereas the old inline path ran it whenever 0 devices were configured. This is a (good) behavior change that respects the opt-out — just calling it out as intentional.

### Blocking: branch is based on pre-#32 `master` and doesn't build standalone
- `merge-base(master, this branch)` is `f6633385` — i.e. **before #32**. This branch's commit absorbed #32's `daemon-server.ts` edits (`shouldConcedePortToOccupant` usage, `pid: process.pid`, the hardened race block) but **not** #32's `session-registry.ts` helper definition.
- So `bridge/src/session-registry.ts` on this branch has **0** definitions of `shouldConcedePortToOccupant`, while `daemon-server.ts` calls it → the branch fails `tsc` on its own. CI only passes because the PR job builds the *merge-with-master* (which supplies the helper from #32).
- Side effect: this PR's diff redundantly re-shows the entire #32 hardening, which muddies the review/history.

### Fix
`git fetch origin && git rebase origin/master` from the branch. After the rebase, #32's `daemon-server.ts` hunks are already in the base (identical patch → dropped), the helper is present, and the diff collapses to **only** the iDotMatrix change. Then it builds standalone and is clean to merge.

PR #52, opened 2026-07-09
feat(sessions): --weight sort override to pin deck/tab order

doug-w:
@
## What this adds

A new **`--weight <n>`** flag on the session commands (`claude`, `codex`, `opencode`, `monitor`) that lets you **pin where a session lands** in the Stream Deck deck — and in every other dashboard surface.

The motivating use case: running several tabs on the same project in Windows Terminal (or iTerm at work). Previously the deck sorted by agent type → project → start time, which is stable but doesnt line up with your terminal tabs. Now you can tag each tab with a weight and the deck mirrors your tab order:

```bash
agentdeck claude --weight 1   # tab 1 → deck slot 1
agentdeck claude --weight 2   # tab 2 → deck slot 2
agentdeck claude --weight 3   # …and so on
```

## How the ordering works

`sortSessions` gains a new **primary** sort key: **weight, ascending**.

1. **weight** ascending — all negatives first, then unweighted/`0`, then positives (`-5, -3, 0, 1, 2, …`)
2. then the *existing* keys unchanged, **within the same weight band**: agentType (openclaw → claude → codex → opencode → antigravity) → project name (natural/numeric) → `startedAt` → id

So weight lets you carve sessions into ordered bands, and everything inside a band keeps sorting exactly as it does today. Negative weights float a session above the unweighted pack (e.g. `--weight -5` to always keep something at the top); positive weights sink it below.

`--weight` is validated as an integer — `--weight foo` fails loudly rather than silently becoming `0`.

## ✅ No weight specified ⇒ **identical to before**

This is the key compatibility guarantee. Weight **defaults to `0`**, and a missing/`null`/non-finite weight is normalized to `0` (`sessionWeight()` on every platform). If you never pass `--weight`, **every** session shares weight `0`, the weight comparison is always a tie, and the sort falls straight through to the original `agentType → project → startedAt → id` logic — **byte-for-byte the same ordering as before this PR**. Nothing about how a session *behaves* changes either; weight is a pure sort key. There is a dedicated test on each platform asserting "no weights == legacy ordering."

## Data flow

```
CLI --weight
  → SessionOptions.weight           (bridge/src/index.ts)
  → SessionEntry.weight             (persisted in sessions.json)
  → EnrichedSession.weight          (daemon aggregation)
  → SessionInfo.weight              (WS wire payload)
  → sortSessions / compareEntries   (every consumer)
```

## Cross-platform parity (hand-mirrored SSOT)

The sort is the shared SSOT in `shared/src/session-utils.ts`, hand-mirrored per the repo convention:

- **shared** — `sessionWeight()` helper + weight as key 0 in `sortSessions`; `SessionInfo.weight`
- **bridge** — `--weight` flag (integer-validated `parseWeight`) on all four session commands; registry + aggregator plumbing
- **Stream Deck plugin** & **TUI dashboard** — inherit automatically (both already re-sort via shared `sortSessions`)
- **Apple (macOS/iOS)** — `DashboardDataRules.sessionWeight` + weight in all three `sortSessions` overloads (incl. the dict-payload path), `DaemonSessionEntry.weight`, daemon emits `weight` on the wire, `SessionListPanel.sortEntries`
- **Android** — `SessionInfo.weight`, `sessionWeight` + `compareSessionsForDisplay`, plus the e-ink `EinkAgentColumn` and tablet `SessionListPanel` `compareEntries`

## Tests

- **shared (vitest)** — 9 new cases: band ordering, `sessionWeight` normalization (null/NaN/Infinity → 0), weight-beats-agentType, within-band fallback, terminal-tab→slot mapping, and no-weight == legacy ordering. Full suite green (**1702 passed**).
- **Apple (XCTest)** — 2 mirror tests in `ProtocolTests.swift`
- **Android (JUnit)** — 3 mirror tests in `SessionDisplayOrderingTest.kt`

## Docs

- `README.md` — new "Pinning session order with `--weight`" section with the 5-tab example
- `CLAUDE.md` — Session sort weight note in the CLI reference

## Notes

- Apple/Android native builds are out of scope on the Windows dev box, so those ports are hand-mirrored and covered by mirror tests; CI validates compilation.
- `generated/protocol/` was intentionally **not** regenerated here — the local `generate-protocol` toolchain produced unrelated churn and didnt add the field to the consumed `.swift`/`.kt`; the app uses the hand-written models updated in this PR. Regenerate the reference artifacts on the canonical macOS/CI toolchain if desired.

@


puritysb (changes requested):
This is currently conflicted and the CI failure is actionable: `shared/src/protocol.ts` adds `SessionInfo.weight`, but the generated artifacts under `generated/protocol/` were intentionally omitted. The CI contract runs `pnpm generate-protocol` and fails on the resulting Swift/Kotlin/schema diff, so hand-updating only the consumed models is not sufficient in this repository. Please rebase on current master, run the canonical protocol-generation workflow, commit the generated artifacts, and rerun CI. The generated field should also be reviewed as an integer-compatible wire value even though the generator represents JSON numbers as Double.

puritysb:
Closing this revision because it is now conflicted with master, its CI still fails on omitted generated protocol artifacts, and there has been no update after the requested changes. The session-weight idea remains valid, but the implementation should be resubmitted from current master with `pnpm generate-protocol` output committed and all platform mirrors/CI green. A fresh PR will be easier to review than repairing this stale 19-file branch across the recent unified-version and timeline changes.

PR #54, opened 2026-07-10
feat(bridge): Linux port — systemd daemon autostart + POSIX fixes

doug-w:
**Ports the Node.js bridge to Linux.** The daemon runtime, mDNS (pure-JS `bonjour-service`), Pixoo/ADB, and the PTY flow already worked; the gaps were autostart and a few macOS-flavored fallbacks. Stream Deck's desktop app is unavailable on Linux, so plugin-host steps are intentionally out of scope.

- **`linux-service.ts`**: systemd `--user` unit (`agentdeck-daemon.service`) mirroring `windows-service.ts` — `buildUnitFile` (`Restart=on-failure` ≈ LaunchAgent KeepAlive), `hasSystemctl` probe, install/start/stop/disable over `systemctl --user`, XDG-aware path. Unit-tested (systemctl calls integration-only).
- **`cli.ts`**: `daemon install`/`uninstall` gain a Linux branch (was a hard exit on non-darwin); degrades to a manual-start hint without systemd; installs Codex/OpenCode hooks for parity.
- **`pty-manager.ts` / `check-deps.ts`**: shell fallback `/bin/zsh` → `/bin/bash` when `$SHELL` is unset (matters for the headless daemon unit).
- **`setup.ts`**: checks for a `cc`/`make`/`python3` toolchain instead of Xcode CLT; skips Stream Deck app/CLI steps (unavailable on Linux).
- **Docs**: `daemon.md` Autostart, README "Linux (Bridge + Daemon)", CLAUDE.md "Linux dev setup" pointer.

Verified on Debian 13 / Node 24: `pnpm build` clean, 1696 tests pass incl. the new `linux-service` suite, `node-pty` compiles natively, `agentdeck` CLI runs.


puritysb (changes requested):
The Linux port is well scoped, but the generated user unit hardcodes `WorkingDirectory=~/.agentdeck`. On a fresh install that directory may not exist, causing systemd to fail at the CHDIR step before Node starts. It also ignores the repository-supported `AGENTDECK_DATA_DIR` override. Please either create the selected data directory during install and use the same resolver as the daemon, or remove `WorkingDirectory` if it is not required. Add a unit test covering a fresh/overridden data directory. The current CI does not exercise the systemctl integration path, so this failure would otherwise only appear on a real new Linux install.

doug-w:
Addressed in 2c0b9a9.

- `WorkingDirectory` now resolves through a new `getDataDir()` that mirrors the daemon's own inline resolver (`AGENTDECK_DATA_DIR` override, else `~/.agentdeck`), so the unit points at wherever the daemon actually reads/writes rather than a hardcoded `~/.agentdeck`.
- `installUnit()` now `mkdir -p`s that data dir before writing the unit, so systemd can't fail at the CHDIR step before Node starts on a fresh (or overridden) install.
- Added unit tests covering the default fallback, the `AGENTDECK_DATA_DIR` override reflected in `WorkingDirectory`, `getDataDir()` resolution, and `installUnit()` creating a not-yet-existing nested data dir before it reaches systemctl.

Full suite green (1180 bridge tests).

puritysb (changes requested):
Thanks for fixing the fresh-directory CHDIR failure. The directory is now created before systemctl runs, and the focused tests/typecheck pass, but the `AGENTDECK_DATA_DIR` override still does not survive into the service process.

`buildUnitFile()` uses the install command's current environment to write `WorkingDirectory=<override>`, but the generated unit does not include an `Environment=AGENTDECK_DATA_DIR=...` directive (or an EnvironmentFile). A systemd user manager is not guaranteed to inherit the interactive shell environment used for `agentdeck daemon install`, and after login/boot it normally will not have this ad-hoc variable. The daemon then executes with no `AGENTDECK_DATA_DIR`, so `session-registry.ts`/APME/etc. fall back to `~/.agentdeck` while systemd merely chdirs into the override directory. This breaks the claimed data-dir parity and can split daemon state across two locations.

Please persist the selected override into the unit with correct systemd escaping (only when explicitly set), or remove override support from this unit and avoid claiming it. Add a test that parses the generated unit and verifies the launched process receives the override, not only that `WorkingDirectory` and the directory exist. It would also be good to quote/escape all generated path values consistently while touching the unit builder.

Verification at 2c0b9a9: `tsc --noEmit` passed and `linux-service.test.ts` passed 12/12.

puritysb:
This remains the closest external PR to merge, so I am keeping it open. The remaining blocker is still the systemd process environment, not only its working directory: when `AGENTDECK_DATA_DIR` is selected during install, the generated user unit must persist that value with correct systemd escaping so the daemon process reads and writes the same directory after login/reboot.

Please add a unit-file test that verifies the launched process receives the override, rebase onto current master (the branch is now 133 commits behind), rerun the full CI matrix, and repeat the fresh Debian systemd user-service install test. Once those are green, this can return for final review.

puritysb:
Friendly ping — it's been a week since the 2026-07-18 review and there are no new commits on `feat/linux-port` (last push 2026-07-12). This is still the closest external PR to merging, so I'd like to land it.

The one open blocker is unchanged: when `AGENTDECK_DATA_DIR` is selected during install, the generated systemd **user unit** must persist that value with correct systemd escaping so the daemon process reads and writes the same directory after login/reboot — not just its `WorkingDirectory`. Please add a unit-file test that verifies the launched process actually receives the override.

Note the branch has fallen even further behind master since the review (recent merges landed), so a rebase onto current master is needed before CI can go green. Please rebase, add that test, rerun the full CI matrix, and repeat the fresh Debian systemd user-service install test.

To keep the queue accurate: if there's no update by **2026-08-01**, I'll close this as stale — no hard feelings, and you're very welcome to reopen or resubmit whenever you pick it back up.

puritysb:
Status update, since you have been active on #53, #62, #81, #82 and #83 today but this one has not moved since 2026-07-12.

**The stale-close deadline I mentioned on 2026-07-25 is off.** You have clearly re-engaged with the queue, so I am not closing this on 08-01 — it remains the external PR closest to merge and I would rather land it.

**Two things it needs.**

1. **A rebase.** GitHub cannot update this branch automatically — it conflicts with current master. Master's tip also had two drifted gates until an hour ago (`sync-hardware-spec-cards` and the `session-utils.ts` SYNC-HASH pin, the ones you correctly diagnosed in #83); both are fixed and master's CI is green, so a rebase now gives you a clean base to run against.

2. **The one remaining blocker from the 2026-07-18 review**, unchanged: a unit-file test proving the generated systemd user unit persists `AGENTDECK_DATA_DIR` *with escaping intact* and that the launched process actually receives the override. The value is user-controlled and travels through a file format with its own quoting rules, so the assertion that matters is on the generated unit's bytes and on the environment the service sees — not on the code path that composes the string.

Everything else in the Linux port read fine on the 07-18 pass.

doug-w:
Rebased onto current master (`334c355d`) and addressed the remaining blocker. Point-by-point against the 2026-07-18 review and the follow-ups:

### 1. Override persists into the service process — `ad90271c`

`buildUnitFile()` now emits an `Environment="AGENTDECK_DATA_DIR=…"` directive, **only when the override is explicitly set** (unset/empty → no `Environment=` line at all). `WorkingDirectory=` and `Environment=` both derive from a single `dataDirOverride` value (the old independent `workingDir` opt is gone), so the unit can never chdir into one directory while the daemon resolves another.

### 2. Consistent quoting/escaping of all generated values — `ad90271c`

Escaping is per-directive, because the rules genuinely differ:

| Directive | `%` | `$` | `\` `"` | quoting |
|---|---|---|---|---|
| `Environment=` | `%%` (specifiers expand) | **single** (not expanded there) | C-escaped | double-quoted |
| `ExecStart=` (both words now quoted, incl. the node path) | `%%` | **`$$`** (expanded even inside quotes) | C-escaped | double-quoted |
| `WorkingDirectory=` | `%%` | literal | **verbatim** (no unquoting — quotes would become path bytes) | none |

The override is also validated before any side effect (absolute path required; control chars, edge whitespace, `~`, and leading `-` rejected with actionable errors — those values are unrepresentable or mean something else in unit syntax), so a rejected install no longer leaves a stray directory.

### 3. Test that the launched process receives the override — `ad90271c`

Two layers in `linux-service.test.ts`, per your "unit's bytes + environment the service sees" framing:

- **Byte-level golden fixtures** pin the exact generated lines for hostile values (space, `%`, `"`, `\`, `$`, incl. a `$`-stays-single-in-`Environment=` guard), plus throw-cases for invalid overrides. The expected bytes encode *observed systemd behavior*, not a docs reading — provenance below.
- A secondary round-trip test parses the generated `Environment=` line with systemd's documented semantics and requires the raw value back (explicitly secondary, since it shares authorship with the escaper).

### 4. Fresh Debian systemd user-service install test — real run

Debian 13 (trixie), systemd 257 (`257.9-1~deb13u1`), from this branch's build. The override directory was deliberately hostile — `…/deck %i "a\$b` (space, specifier `%i`, embedded `"`, `\`, `$`):

```
override dir: …/scratchpad/hostile.0joK/deck %i "a\$b

-- generated unit (relevant lines) --
ExecStart="…/bin/node" "…/bridge/dist/cli.js" daemon start --foreground
WorkingDirectory=…/hostile.0joK/deck %%i "a\$b
Environment="AGENTDECK_DATA_DIR=…/hostile.0joK/deck %%i \"a\\$b"

PASS: unit bytes: exact Environment line
PASS: unit bytes: exact WorkingDirectory line
PASS: systemctl show contains AGENTDECK_DATA_DIR (advisory)
PASS: /proc/<MainPID>/environ AGENTDECK_DATA_DIR == raw override (byte-exact)
PASS: process cwd == override dir
PASS: daemon.json written inside the hostile override dir
PASS: no daemon.json leaked into ~/.agentdeck
PASS: override survives systemctl restart (new PID)
PASS: uninstall removed the unit

==== PASS 2: fresh default (no override) ====
PASS: default unit has NO Environment= line
PASS: default WorkingDirectory=~/.agentdeck
PASS: fresh install: ~/.agentdeck created + daemon.json written
PASS: fresh pass: uninstall removed the unit

==== RESULT: 13 passed, 0 failed ====
```

The equality assertion is on `/proc/<MainPID>/environ` (raw, NUL-delimited) and on where `daemon.json` lands — `systemctl show` is treated as advisory since it prints its own re-escaped serialization. These observed bytes are what the golden fixtures pin (noted in a provenance comment in the test file).

### 5. Rebase notes

Master's README rewrite moved platform detail into `docs/` (`docs/windows.md`), so the PR's "Linux (Bridge + Daemon)" README section was re-homed as **`docs/linux.md`** and linked from the README the same way as Windows. `setup/src/setup.ts` was re-applied onto the restructured prerequisite checks and follows master's newer soft-check philosophy: the Linux toolchain check **warns** instead of failing (matching the prebuilt-first `installBridge`, whose source-build fallback now prints distro toolchain hints on Linux instead of `xcode-select --install`), and the Stream Deck app check is skipped on Linux. `agentdeck daemon install` now also logs the resolved data dir and whether the override was persisted; `docs/daemon.md` documents that the override is captured at install time and refreshed by re-running install.

(The design-system catalog exclusion for `docs/linux.md` — same documented reason as `docs/windows.md` — ships inside the port commit itself, so every commit in the series passes the Design System gate.)

Full suite green after rebase and again after the fix (2076 tests, 121 files); `tsc --noEmit` clean; both PR checks green at `ad90271c`.

puritysb (approved):
기존 변경 요청의 차단 조건이 해소됐음을 확인했습니다. `AGENTDECK_DATA_DIR`는 단일 override source로 정리됐고, systemd directive별 escaping(`Environment=`, `ExecStart=`, `WorkingDirectory=`)과 입력 사전 검증이 추가됐습니다. 작성자가 Debian 13/systemd 257 환경에서 hostile path를 포함해 `/proc/<pid>/environ`, cwd, daemon.json 경로, 재시작 및 제거를 실제 검증한 결과도 확인했습니다.

최신 master를 head `45a7cf2d5a863a80e01c0617743536ff4e732c69`에 통합한 뒤 로컬에서 version/build/docs/mirror 및 전체 테스트(124 files, 2,107 tests)를 통과했고, GitHub CI도 모두 통과했습니다.

PR #55, opened 2026-07-10
fix(bridge): correctly render AskUserQuestion / permission / diff prompts on the deck

doug-w:
## Summary

A series of related fixes to the Claude Code PTY output parser (`bridge/src/output-parser.ts`) so interactive prompts — `AskUserQuestion`, permission approvals, diff/edit approvals — render correctly on the Stream Deck / D200H instead of showing the wrong options, a stale option, or a false **idle**. Each was root-caused from a live `-d` capture and locked with a regression test.

## Commits / fixes

**1. Parse all `AskUserQuestion` options (`e3788fd3`)**
An `AskUserQuestion` prompt (e.g. `a`, `b` + Claude's `Type something.` / `Chat about this` affordances) rendered with the real leading options **dropped** — the trailing affordances got renumbered as slots 1–2. Two compounding causes: the descriptions render **unindented** (col 0), so the option-block backward scan broke on them and cut off the leading options; and a full-width `────` rule between options got its newline stripped, gluing hundreds of box chars onto the adjacent label and exhausting the scan window. Fix: tolerate an unindented description **only when an option sits directly above** (still excludes stale numbered lists), collapse oversized box-drawing rules at buffer-build time, widen/name the option-scan window, and gate the 0-based re-index to genuine truncation. Also made the debug-log path portable (`os.tmpdir()`), since the hardcoded `/tmp` silently wrote nothing on Windows.

**2. Stale numbered prose can't overwrite the `❯` option (`d8bd37b4`)**
A markdown `1. …` line from the assistant's own chat scrollback could be scanned into the option block and, mapping to index 0 with no cursor, **overwrite the real `❯ 1. Yes`** — so a Yes/No prompt rendered as `[<prose>, No]`. Fix: a `❯`-marked option is authoritative for its index; a later non-cursor line at the same index can no longer clobber it.

**3. Show AWAITING when a prompt coincides with the spinner (`4ed10b06`)**
Claude Code animates a status spinner (`✻`/`✽`) at the same time a permission/option prompt is on screen. Each spinner chunk was cancelling the pending option debounce, so the prompt never emitted and the deck sat at idle/processing instead of AWAITING. Fix: ignore a spinner char while a `❯`-marked navigable prompt is on screen (the `❯` distinguishes a real prompt from a numbered list in prose; the "spinner cancels option timer" bare-list case is untouched). An active spinner with no prompt still reads as working, never idle.

**4. Keep AWAITING when a spinner floods the buffer after a prompt (`ed07a415`)**
The remaining case: a diff/edit approval that appears while a build spinner runs still went idle. The `✻/✽` status line redraws every ~100ms and floods the linear buffer, pushing the `❯` options past the buffer window within a few hundred ms → the spinner cancels the option debounce → the state machine flips AWAITING → processing → idle. The buffer-tail check can't win that race (the linear buffer doesn't reflect screen position), so a **time-based backstop** was added: arm a suppression window when a *real* prompt is detected (`❯`-marked / permission / diff — never a bare numbered list), and for ~1.5s treat spinner/idle signals as the concurrent animation (don't cancel the debounce, don't emit idle). Re-armed from the buffer so a long-lived prompt keeps suppressing; the navigable-prompt window was also widened 600 → 2000.

## Tests

Each fix ships a regression test in `bridge/src/__tests__/output-parser.test.ts` reproducing the exact captured shape (unindented descriptions + giant rule; a stale `1.` prose line below a `❯` prompt; a prompt with concurrent spinner chunks; a prompt followed by 80 spinner status redraws). The existing prompt-shape and "spinner cancels option timer / stale numbered list" guards stay green. Full `output-parser.test.ts` (219) and full bridge suite (1171 / 2 skipped) pass.

## Verification

All four were reproduced and fixed end-to-end on a live `-d` session driving a Stream Deck — the deck now shows the real options and reaches AWAITING for `AskUserQuestion`, Yes/No permission, and diff/edit approvals, including while a build spinner is running.

### DIFF
--- a/bridge/src/__tests__/output-parser.test.ts
+++ b/bridge/src/__tests__/output-parser.test.ts
@@ -2423,6 +2423,33 @@ describe('OutputParser', () => {
       expect(events[0].options).toHaveLength(4);
     });
 
+    it('a stale markdown "1." line from chat must not overwrite the ❯-marked option', () => {
+      const p = armParser();
+      const events: any[] = [];
+      p.on('option_prompt', (d) => events.push(d));
+      p.on('permission_prompt', (d) => events.push(d));
+
+      // Reproduces agentdeck-debug.log: a Bash approval prompt (❯ 1. Yes / 2. No)
+      // with a markdown numbered line from the ASSISTANT'S OWN chat message landing
+      // in the same scanned block. The stale "1." maps to index 0 with no cursor and
+      // must NOT clobber the real "❯ 1. Yes" — the ❯ cursor is authoritative.
+      // Long numbered prose line with the same shape as the captured case — keeps
+      // the `>` and `( )` characters (so label cleaning is still exercised) without
+      // embedding real conversation text.
+      p.feed([
+        'Do you want to proceed?',
+        '❯ 1. Yes',
+        '  2. No',
+        '',
+        '1. Lorem ipsum dolor sit amet > consectetur (adipiscing elit) sed do eiusmod tempor incididunt ut labore et dolore magna aliqua ut enim ad minim veniam.',
+      ].join('\n'));
+      vi.advanceTimersByTime(200);
+
+      expect(events.length).toBeGreaterThanOrEqual(1);
+      const opts: PromptOption[] = events[events.length - 1].options;
+      expect(opts.map((o) => o.label)).toEqual(['Yes', 'No']);
+    });
+
     it('still works with scrambled TUI order after backward scan', () => {
       const p = armParser();
       const events = collectEvents(p, 'option_prompt');
--- a/bridge/src/__tests__/output-parser.test.ts
+++ b/bridge/src/__tests__/output-parser.test.ts
@@ -2741,6 +2741,51 @@ describe('OutputParser', () => {
     });
   });
 
+  // === AskUserQuestion tall prompt: full-width rule must not eat leading options ===
+  //
+  // Regression for the live bug captured in a `-d` session (agentdeck-debug.log,
+  // lines 2617-2620): AskUserQuestion draws a full-width horizontal rule between the
+  // trailing affordances. ANSI cursor-positioning strips its newline, so the rule
+  // (measured ~270 "───" triples) is glued onto the "Type something." label AND
+  // consumes the whole option-scan window, so the real leading options (a, b) fall
+  // outside `slice(-N)` and are never parsed — the trailing affordances then get
+  // re-indexed to slots 1-2 and `navigable=false` is emitted (stuck AWAITING).
+  describe('AskUserQuestion tall prompt with oversized separator', () => {
+    it('keeps leading options a/b when a giant box-drawing rule contaminates a later option', () => {
+      const p = armParser();
+      const events = collectEvents(p, 'option_prompt');
+
+      // Faithful to the live capture (agentdeck-debug.log DIAG dump): the giant
+      // rule is glued to option 3's label with no newline, AND the per-option
+      // descriptions render UNINDENTED (col 0) — both conditions must be handled.
+      const rule = '─── '.repeat(1000); // ~4000 chars
+
+      p.feed([
+        'Which option do you choose?',
+        '❯ 1. a',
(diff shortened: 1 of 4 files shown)


puritysb (changes requested):
The captured-layout regression is valuable, but the new spinner suppression has a possible first-chunk hole. In the non-active-spinner path, if a chunk contains both a spinner character and a navigable `❯` option prompt, `bufferHasNavigablePrompt()` becomes true and the branch returns immediately. That skips the option-detection code below, so if this is the first prompt chunk no option timer may be armed and AWAITING may still never be emitted. Please add a regression test feeding spinner + navigable AskUserQuestion prompt in the same initial `feed()` call and adjust control flow so spinner suppression does not skip prompt parsing. Also note that spinner/idle policy and logger path changes expand the scope beyond box-rule parsing; keeping tests for each behavior will make this safe to merge.

doug-w:
Fixed in 24018bd0.

**First-chunk hole:** the spinner-suppression branch now falls through to the option/permission detection below instead of `return`ing. Previously, when `bufferHasNavigablePrompt()`/`recentInteractivePrompt()` was true it returned immediately, skipping the option-timer arming — so a chunk carrying *both* a spinner char and the navigable `❯` prompt would never reach AWAITING. The non-suppressed path (real spinner → `spinner_start` + cancel timers + return) is unchanged; only the suppressed branch drops its early return.

**Regression test:** added `emits AWAITING when a spinner char and the ❯ prompt arrive in the SAME initial chunk` — a single `feed('✳ Cooking…\nWhich approach?\n❯ 1. …')`. Verified it fails on the prior commit and passes after the fix. This sits alongside the existing spinner-before-prompt and spinner-floods-after-prompt cases, so each spinner/idle behavior keeps its own test. Full suite: 220 passing, `tsc` clean.

puritysb (changes requested):
The requested same-initial-chunk fix is correct and its regression test passes, but the new buffer-based suppression introduces a stale-prompt processing hole.

`bufferHasNavigablePrompt()` searches the append-only last 2,000 characters, and the spinner branch re-arms `lastInteractivePromptAt` whenever that search finds `❯ <n>`. After the user answers a prompt, the old option text can remain in the buffer even though it is no longer on screen. A subsequent real spinner then gets ignored; because each ignored spinner re-arms the timestamp, repeated processing updates can suppress `spinner_start` indefinitely and leave the deck in AWAITING/idle while the agent is actually working.

I reproduced this with a focused test: arm the parser, feed `❯ 1. Yes / 2. No`, let the option emit, advance 2 seconds beyond `INTERACTIVE_PROMPT_SUPPRESS_MS`, then feed `✻ Running the selected action`. Expected: one `spinner_start`; actual: none. The existing output-parser suite still passes 220/220, which confirms this lifecycle is currently uncovered.

Please avoid treating historical buffer presence as proof that the prompt is still on screen. The suppression needs an explicit prompt lifecycle/freshness signal that is cleared on selection/processing, or at minimum must not re-arm indefinitely from stale buffer contents. Add the answered-prompt → real-spinner regression above while preserving the same-chunk and spinner-flood cases.

puritysb:
I am splitting this PR along the safety boundary identified in review:

- the first two commits (`e3788fd3`, `d8bd37b4`) are independent option-block parsing fixes and apply cleanly to current master; I am carrying them into a focused PR with original authorship preserved;
- the spinner/prompt suppression commits are not being carried because historical append-only buffer content can indefinitely hide real processing after a prompt is answered.

The spinner lifecycle is now tracked independently in #68 with the answered-prompt → real-spinner regression and the requirement for an explicit prompt lifecycle/freshness signal. Once the focused parsing PR is published, this combined PR will be closed as superseded rather than merged wholesale.

puritysb:
Superseded by #71, which preserves the two reviewed parser/logging commits and their original authorship on top of current `master`. It passes all 216 output-parser tests and the full workspace build.

The spinner lifecycle rewrite is deliberately not carried forward; that behavior needs a smaller state-machine design and targeted concurrency regression tests, now tracked in #68. Thank you for the detailed live capture and root-cause work.

PR #62, opened 2026-07-12
feat(sessions): --weight sort override to pin deck/tab order (resubmit of #52)

doug-w:
Resubmits the session-weight feature from #52, rebased onto current master with the requested changes applied. #52 was closed as stale/conflicted with CI failing on omitted generated protocol artifacts; this branch fixes all three concerns.

## What this adds

A `--weight <n>` flag on the session commands (`claude`, `codex`, `opencode`, `monitor`) that pins where a session lands in the Stream Deck deck — and in every other dashboard surface. Running several tabs on one project (Windows Terminal / iTerm), you tag each tab with a weight and the deck mirrors your tab order:

```bash
agentdeck claude --weight 1   # tab 1 → deck slot 1
agentdeck claude --weight 2   # tab 2 → deck slot 2
```

`sortSessions` gains a new **primary** key: **weight ascending** (negatives → unweighted/`0` → positives), then the existing `agentType → project → startedAt → id` keys unchanged *within* a weight band. `--weight` is integer-validated (`--weight foo` fails loudly). Weight defaults to `0`; leaving every session unweighted reproduces the pre-weight ordering byte-for-byte. Weight is a pure sort key — no behavioral change.

## Changes vs #52 (addresses the close reasons)

- **Rebased onto current master** — resolved the `bridge/src/cli.ts` conflict (master removed the `d200h` module key; kept master's module set and threaded `weight`). No other conflicts.
- **`pnpm generate-protocol` output committed** — the CI "Protocol drift check" now passes. Regenerated `generated/protocol/BridgeEvent.swift`, `BridgeEvent.kt`, and `bridge-event-schema.json` add the `weight` field.
- **Integer-compatible wire value confirmed** — the generator represents JSON numbers as `Double`, so the reference artifacts show `weight: Double?`. The consumed hand-written models use `Int?` (`SessionInfo.weight` in Swift/Kotlin) and the daemon emits an integer (`parseWeight` validates as int). JSON `2` decodes cleanly into both `Int` and `Double`, and the Swift dict-payload path coerces via `NSNumber.intValue`. So the reference-vs-consumed type split is intentional and wire-safe.

## Local CI parity

Ran the full `ci.yml` + `design-system.yml` sequence locally before pushing — all green:

- `pnpm verify-version` ✓
- `pnpm build` ✓ · `pnpm typecheck` ✓
- `pnpm test` — **1702 passed**
- `npx vitest run --coverage` — thresholds cleared (lines 42.4%, branches 37.8%, functions 39%, statements 40.9%)
- **Protocol drift check** — `pnpm generate-protocol` produces no diff ✓ (the #52 blocker)
- Token mirror sync + design-lint vs baseline ✓ (89 ≤ 93)

## Data flow

```
CLI --weight → SessionOptions.weight → SessionEntry.weight (sessions.json)
  → EnrichedSession.weight → SessionInfo.weight (wire) → sortSessions / compareEntries
```

## Cross-platform parity (hand-mirrored SSOT)

SSOT in `shared/src/session-utils.ts` (`sessionWeight()` + weight as key 0 in `sortSessions`), hand-mirrored in bridge (`parseWeight` flag on all four commands + registry/aggregator plumbing), Apple (`DashboardDataRules.sessionWeight`, all `sortSessions` overloads incl. dict path, daemon wire emit), and Android (`sessionWeight` + `compareSessionsForDisplay`, e-ink `EinkAgentColumn`, tablet `SessionListPanel`). Stream Deck plugin + TUI inherit via shared `sortSessions`.

## Tests

- **shared (vitest)** — 9 cases: band ordering, `sessionWeight` normalization (null/NaN/Infinity → 0), weight-beats-agentType, within-band fallback, tab→slot mapping, no-weight == legacy ordering
- **Apple (XCTest)** — 2 mirror tests in `ProtocolTests.swift`
- **Android (JUnit)** — 3 mirror tests in `SessionDisplayOrderingTest.kt`

Apple/Android native builds are out of scope on this dev box; those ports are hand-mirrored and covered by mirror tests. CI validates compilation.


puritysb:
Thanks for resubmitting this with the generated protocol artifacts and green CI. GitHub currently reports the PR as `mergeable: false`, `mergeable_state: dirty`, and `rebaseable: false` against the latest master, which has advanced since this branch was prepared. Please rebase/recreate the branch on current `puritysb/AgentDeck` master, resolve the conflicts without dropping the recent daemon/protocol/platform changes, and rerun the full CI matrix. I will do the substantive weight-ordering review on the conflict-free head so we do not review code that cannot be merged as submitted.

puritysb (changes requested):
Thanks for updating the branch. The new head is clean/mergeable and all four CI checks pass. I completed the substantive review and found several cross-surface gaps that need to be fixed before merge:

1. **Weight is lost on the Swift-daemon discovery path.** `startSession()` persists `weight` in the Node `SessionEntry`, but `DaemonWsClient` does not include it in `session_push_register`, and Swift `handleSessionPushRegister` neither reads nor preserves it. The Swift daemon cannot read the CLI process's `~/.agentdeck/sessions.json`; its own comment identifies this push registration as the sole discovery path. Consequently `--weight` does not reach the App Store macOS/iOS dashboards when the built-in Swift daemon is active, despite the added Swift model/sort fields. Please thread weight through the register frame and preserve it during Swift re-register/state entry reconstruction, with a regression test for the Node-session → Swift-daemon shape.

2. **Weighted Codex tabs are folded before they can be sorted.** Stream Deck and D200H call `foldCodexSessionsForDisplay()` before `sortSessions()`. The fold key is only Codex kind + project, so two same-project Codex tabs with weights 1 and 2 collapse to one representative whose weight can change with state. I reproduced this in a focused test: expected `[codex-tab-1, codex-tab-2]`, actual `[codex-tab-2]`; the existing 34 session-utils tests still pass. Explicitly distinct weights need to participate in the fold key or otherwise opt those sessions out of folding, mirrored in the Swift folding implementation/preview. Add a regression covering two weighted same-project Codex sessions and state changes.

3. **The CLI accepts integers that are not wire-safe and the native comparators can overflow.** `Number.isInteger(Number(value))` accepts values such as `1e100` and `9007199254740992`; even `2147483648` cannot fit the Kotlin `Int?` model. Kotlin and Swift also subtract user-controlled weights, so opposite/extreme values can overflow (Swift may trap; Kotlin can reverse ordering). Please enforce a documented cross-platform range in `parseWeight`, use direct/three-way comparison instead of subtraction in TS/Swift/Kotlin, and test both bounds plus opposite-sign extremes.

4. **Required mirrors/docs are incomplete.** `D200HLayoutModel.D200HSession` and its hand-ported `sortSessions` still have no weight, so Device Preview no longer mirrors the shared layout ordering. Because this adds behavior to App Store dashboard surfaces, add the required `docs/appstore-feature-matrix.md` row, and record the structural CLI→registry→wire→surface change in `DEVELOPMENT_LOG.md` rather than relying only on the CLAUDE/README text.

Verification at b27731d3: shared/bridge TypeScript typechecks passed; existing session-utils tests passed 34/34; the focused weighted-Codex-fold regression failed as described; GitHub CI is green.

puritysb:
Moving this PR back to Draft to match the unresolved cross-surface contract. The feature remains useful, but the current head now conflicts with master and still needs all four substantive review areas addressed:

1. carry weight through `session_push_register` into the Swift daemon and preserve it on re-registration;
2. prevent distinct weighted same-project Codex sessions from being folded together;
3. enforce a documented cross-platform integer range and replace subtracting comparators with overflow-safe comparison;
4. mirror D200H preview ordering and update the App Store feature matrix plus DEVELOPMENT_LOG.

Please rebase after implementing these changes, regenerate protocol mirrors, run the full TypeScript/Apple/Android test matrix, and only then mark ready for review.

puritysb:
Friendly ping — no new commits since 2026-07-12 and the 2026-07-18 draft review still stands. The current head conflicts with master (now even further behind after recent merges).

The four substantive review areas remain unaddressed:
1. carry weight through `session_push_register` into the Swift daemon and preserve it on re-registration;
2. prevent distinct weighted same-project Codex sessions from being folded together;
3. enforce a documented cross-platform integer range and replace subtracting comparators with overflow-safe comparison;
4. mirror D200H preview ordering and update the App Store feature matrix plus DEVELOPMENT_LOG.

Please rebase after implementing these, regenerate the protocol mirrors, run the full TypeScript/Apple/Android test matrix, and only then mark ready for review. If there's no update by **2026-08-01**, I'll close this as stale to keep the queue clean — you're very welcome to resubmit afterward.

doug-w:
Rebased onto current master (head is a single commit on today's tip) and implemented all four review areas. Point-by-point:

**1. Weight through `session_push_register` + preserved on re-registration**
- `bridge/src/daemon-ws-client.ts` — `SessionPushRegister` gains `weight`, and the sender **always includes a concrete integer** (unweighted ⇒ explicit `0`, never omitted). Receivers merge retain-on-absent, so an only-sent-when-set key could be set but never reset — the same one-way-latch class as the `usageStale` bug this repo already documents.
- Swift daemon — the register merge is extracted into `DaemonServer.mergedPushRegisterEntry` (nonisolated static, XCTest-covered in `SessionPushRegisterMergeTests`): explicit value > absent > retain, clamped on ingestion. Weight is carried through **every** entry-reconstruction site: the re-register recreate, the `handleSessionPushState` projectName recreate (the same field-by-field rebuild that once wiped state/modelName/effortLevel), and the `session_start` hook handler's same-id overwrite. `sessionToDict` emits weight clamped at that single choke point, which also covers the dev-build daemon's direct `sessions.json` Codable ingestion.
- Node daemon receiver is intentionally unchanged: it only acks push-register and discovers sessions via `sessions.json`, which already carries weight through the aggregator. The other hook-path recreates (`codex_session_start`, etc.) are unreachable for weighted sessions — their hooks route to the bridge via `AGENTDECK_PORT` — so they were left alone rather than churned.
- Regression tests: vitest on the register frame shape (`daemon-ws-client.test.ts` — explicit-0 default, set value, garbage normalized) and XCTest on the merge rule (fresh register, legacy-sender preserve, recreate carry-through, explicit-0 reset).

**2. Distinct weighted same-project Codex sessions no longer fold**
The fold key now includes the normalized weight band — in **all three** fold implementations, not just the shared one: `shared/src/session-utils.ts`, the Swift daemon's server-side `DashboardDataRules.foldCodexSessionPayloadsForDisplay` (which runs *before* `sortSessionPayloads` in the `sessions_list` build, so a weight-blind key discarded a pinned tab **on the wire** where no client could recover it), and the `D200HLayoutModel` preview hand-port. Unweighted and equal-weight sessions fold byte-identically to before. Your reproduction is now a regression test in TS (`[codex-tab-1, codex-tab-2]` survives, including across state flips) and both Swift folds.

**3. Documented range + overflow-safe comparison**
- Range is **-9999..9999**, SSOT `SESSION_WEIGHT_MIN/MAX` in `shared/src/session-utils.ts`, emitted to Swift/Kotlin by a new `scripts/generate-session-weight-rules.mjs` following the `generate-terrarium-rules` pattern (generated `SessionWeightRules.generated.swift/.kt` + vitest drift gate that imports the emitters — no new hand mirror; the deliberately self-contained `D200HLayoutModel` copy is grep-gated by the same test as acknowledged hand-port debt). Note: `project.pbxproj` picks up the generated Swift file via CI's `xcodegen generate`; xcodegen isn't runnable from this Windows box.
- `parseWeight` rejects non-integers and out-of-range with the range in the message (`1e100`, `9007199254740992`, `10000` all fail loudly).
- Enforcement is layered because CLI-parse alone isn't enough: the Node aggregator sanitizes at ingestion (`sanitizeWeightForWire` — a stale-CLI or hand-edited `sessions.json` value like 2^40 would otherwise be re-emitted verbatim, and kotlinx's `Int?` decode then aborts the **entire** `sessions_list` event, freezing the Android dashboard before any client clamp could run), the Swift daemon clamps at ingestion and emission, and display-side `sessionWeight()` normalizes (trunc + clamp) in all mirrors as defense in depth.
- Every comparator is now three-way (`compareTo` / `<`), never subtraction: TS `sortSessions`, both Swift `sortSessions` overloads + `SessionListPanel.sortEntries` + the D200H port, and all three Kotlin comparators. Tests cover both bounds, bound±1, and opposite-sign extremes (`Int.MIN_VALUE` vs `Int.MAX_VALUE` in Kotlin — the case where subtraction wraps and reverses the ordering).

**4. D200H preview mirror + docs**
- `D200HSession` carries `weight`; the hand-ported `sortSessions` gains the weight key first (three-way) and the fold-key change; the `liveD200HInput` `SessionInfo → D200HSession` mapping passes weight through (this was the silent gap — without it every live preview session got `nil` and the preview still diverged); the snapshot test asserts the mapping and the resulting order. The `session-utils.ts` SYNC-HASH pin is refreshed and `check-preview-mirror-sync` passes.
- `docs/appstore-feature-matrix.md` gains the row under Agent sessions as **Limited** (weight is CLI-set at launch; a pure App Store install's observed-hook sessions never carry it — matching the house style of the terminal-session-discovery row). `DEVELOPMENT_LOG.md` records the structural CLI→registry→wire→surface chain, the fold-key decision, and the range. The range is documented in `docs/cli.md` (the CLI reference moved there when the README became an entry point) and CLAUDE.md.

**Test matrix, honestly stated**
- TypeScript: `pnpm build` / `typecheck` green; vitest 2023 passed locally — the only failures are this Windows dev box's pre-existing environmental ones (identical set reproduces on origin/master: CRLF script-spawn in `protocol-generated-sync`/`terrarium-rules`, the PowerShell hook variant in `install.test.ts`, tag resolution in `version-policy`). Protocol drift: regeneration produces no diff.
- Apple: not runnable natively here, but `apple-test.yml` runs `xcodebuild build` + `test` on this PR (it triggers on `apple/**`, `shared/**`, `generated/protocol/**`), so the Swift changes — including the new `SessionPushRegisterMergeTests` — are validated pre-merge in CI.
- Android: attempted `gradlew testDebugUnitTest` locally (JDK 21 present, Gradle configures) — it fails at `SDK location not found`; there is no Android SDK on this box. And to be precise: no PR-triggered workflow compiles Kotlin (`test-report.yml` is push-to-master only), so the Kotlin changes and the extended `SessionDisplayOrderingTest` first execute in CI after merge. They are small hand-mirrors of the tested TS/Swift logic, and the cross-language literal drift gate runs in vitest on this PR.

**CI on this head:** `test` (vitest incl. the protocol drift check and the new weight drift gates), `test-macos` (xcodebuild test — runs the new XCTests), `build-ios`, and `validate-submission` are green. The one red check, **Token sync + lint regression**, is inherited: it fails identically on master's own tip (`de9bdb6c`, the LilyGo companion UX commit — `sync-hardware-spec-cards.mjs --check` reports `docs/hardware/index.html` out of sync, and master's design-system run at that commit shows the same failure). Nothing in this PR touches the hardware docs.

Since that inherited check keeps the head from going fully green, I'm leaving the PR in Draft for now — once master's spec-card sync is fixed I'll rebase and mark it ready. All four review areas are implemented and CI-validated on this head in the meantime.

puritysb:
This is a thorough response, and the point-by-point structure made it reviewable in one sitting — thank you. All four areas check out. I also closed the one gap you flagged honestly.

**I ran your Android tests.** JDK 17 + Android SDK on this machine: `compileDebugKotlin` builds clean and `SessionDisplayOrderingTest` reports **11 tests, 0 failures**, including all five new weight cases — null→0, the documented bounds and clamping beyond them, `Int.MIN_VALUE` vs `Int.MAX_VALUE` ordering, weight-before-agentType, and the five-tab pin. So the Kotlin hand-mirrors are no longer unexecuted; that item in your test matrix is closed.

On the design, area by area:

1. **Explicit `0` in the register frame** is exactly the right call, and citing the `usageStale` precedent is the correct reading of it: the failure mode is not "the flag is missing", it is "the flag can be set but never reset". Extracting the merge into `mergedPushRegisterEntry` as a nonisolated static with its own XCTest is better than the alternative, because the bug class here is a field-by-field entry rebuild silently dropping a field — a single choke point is what stops the next one.
2. **Weight in the fold key.** Fold-runs-before-sort was the part I wanted verified, and you found the worse instance yourself: the Swift daemon folds while building `sessions_list`, so a weight-blind key discarded a pinned tab *on the wire*, where no client could recover it. Covering all three implementations rather than only the shared one is the right scope.
3. **Range and comparators.** A generated SSOT with a drift gate is the correct answer here, not a fourth hand mirror. Layering the enforcement — CLI parse, aggregator ingestion, Swift emission choke point, display-side normalize — is justified by the kotlinx `Int?` failure you describe: one out-of-range number aborting the whole `sessions_list` decode is a much bigger blast radius than a mis-sorted row, and a clamp that only runs client-side would never get the chance.
4. **D200H preview.** The `liveD200HInput` mapping was the real gap; a mirror that compiles but receives `nil` for the new field is the kind of drift a SYNC-HASH pin does not catch on its own.

**Two small things, neither blocking:**

- `DaemonWsClient`'s constructor gained `weight` as a positional parameter inserted *before* `resolveDaemonPort`. Typecheck catches a missed call site today, but it is a fragile signature to keep growing, and PR #53 also edits this file — whichever of the two lands second will conflict right here. An options object would age better if you are touching it anyway.
- `sanitizeWeightForWire`'s doc comment says out-of-range values are kept "out of `sessions_list` entirely", but the implementation clamps them rather than omitting them. The behavior is the right one; the comment just overstates it.

**The inherited red check is fixed.** Master's tip had drifted on two gates — `sync-hardware-spec-cards` (the T-Display-S3-Pro Input rows you correctly identified in #83) and the `session-utils.ts` SYNC-HASH pin in `D200HLayoutModel.swift`. Both are resolved on master and its own CI is green.

Please rebase. GitHub cannot update this branch automatically because you will hit two conflicts, both trivial:
- `D200HLayoutModel.swift` — master updated the same `session-utils.ts` SYNC-HASH pin to `ae936f86…`, byte-identical to yours, so either side resolves it.
- `DEVELOPMENT_LOG.md` — both sides add an entry at the top; keep both.

Once that is pushed and the checks are green, mark it ready for review and I will merge.

puritysb (approved):
Final pass complete on the latest head after integrating npm 1.0.4/master. The only merge conflict was the CLI import block; both the session-weight SSOT imports and the new BLE runtime imports are preserved.

The requested-change areas are closed: explicit-zero reset semantics survive session_push_register rebuilds, weight participates in every pre-sort Codex fold, the generated -9999..9999 contract is enforced with overflow-safe comparisons, and Swift/Kotlin/D200H mirrors are wired and tested. Local integration verification passed (`pnpm verify-version`, build, mirror/docs checks, 123 files / 2,070 tests), and all five GitHub checks are green on this exact head.

PR #73, opened 2026-07-18
Move the Swift daemon off the main actor

puritysb:
## Why

The macOS daemon is hosted inside the GUI app and was `@MainActor`, so SwiftUI rendering and daemon service shared one executor.

Measured 2026-07-18: a Debug build whose Terrarium animation ran under Metal API Validation + Xcode queue debugging saturated the main runloop (render 83%, idle 0%). The daemon went unreachable — `/health` returned nothing for 5s and the log stalled for **24 minutes** — while the process stayed alive and WebSocket traffic kept flowing.

That asymmetry (HTTP dead, WS alive) is the tell. `WebSocketServer` owns the daemon's only listener and runs it on its own `ioQueue`, so accept and frame I/O kept working. Plain HTTP arrives on that same listener and is delegated to `HTTPServer.handle(request:on:)`, whose route handlers dispatch into `DaemonServer` — `@MainActor`. **Requests were accepted and then starved.**

Nothing in `DaemonServer` needed the main thread: no `ObservableObject`, no `@Published`, no AppKit/SwiftUI imports, zero `NS*`-UI calls. 178 of 179 methods have no UI dependency. The isolation dates from the file's first commit, when it was 894 lines, and was never revisited across 10× growth.

## Commits

1. **`a8f3e27c`** — moves `HTTPServer`'s listener off the main queue. ⚠️ Its message claims this fixed the outage. **It did not** — see 5) — but the change is correct for that path and keeps the two listeners aligned.
2. **`046c5cb1`** — pins the order-sensitive regions *before* touching isolation, so the flip stays mechanical. Correct under both isolations, revertable on its own.
3. **`ac610d4b`** — the flip, onto a `@DaemonActor` global actor. **This is the commit that fixes the outage.**
4. **`ff83c564`** — the property the refactor exists for, as a test, plus architecture docs.
5. **`5d10598e`** — corrects the record: `HTTPServer.start(port:)` has no callers outside tests, so its listener was never involved. Documents the real mechanism where the next reader will hit it.

**Why a global actor and not `actor DaemonServer`**: the daemon is not one object. `StateMachine` (87 call sites) and `ApmeCollector` (18) expose synchronous value-returning methods taking non-Sendable `[String: Any]`. Independent actors would force `await` across ~105 call sites; plain classes have no isolation for `Task {}` to inherit (this is what broke `StateMachine.resetStuckTimer`). One shared global actor keeps those calls synchronous and preserves isolation inheritance exactly as under `@MainActor` — "same semantics, different executor".

`DaemonVoiceAssistant` stays `@MainActor`: AVAudioEngine / SFSpeechRecognizer / AVSpeechSynthesizer are the one place a thread change carries runtime risk the compiler cannot check.

## The bug Swift 6 could not catch

Swift 6 language mode verifies data-race safety. It cannot see executor assumptions in C callbacks.

The IOKit power notification, written inline in `startServices()`, statically inherited `@DaemonActor`. A `@convention(c)` pointer cannot carry isolation, so Swift compiled an executor assertion into its entry — and IOKit calls it on the port's dispatch queue. The app aborted seconds after launch with `Incorrect actor executor assumption`. It now lives at file scope, where there is no isolation to inherit. The other C/dispatch callbacks were swept.

This surfaced only because the app was launched and watched. A green build and a green suite both passed while it was broken.

## Verification

- 453 tests pass
- App runs with zero isolation aborts; `/health` `/status` `/devices` answer in ~1.5ms; ESP32 boards enumerate; hooks flow; 16 WS clients reconnect
- `DaemonActorIndependenceTests` — daemon work completes while the main thread is pinned. Forcing the same body back onto `@MainActor` makes it **time out entirely**, reproducing the starvation on demand

## Not verified

- **Sleep/wake recovery** — the IOKit path just fixed. The app survived a suspend without crashing, but `System wake` never appeared in the log (zero occurrences historically too), so the handler firing is unconfirmed
- **ESP32 connect-burst reentrancy** — needs a board disconnecting mid-burst
- **Hook turn serialization** — drift would show up only over days of use
- **Voice assistant** — callbacks now cross MainActor↔DaemonActor; no automated coverage

## Known, not addressed here

Handing port 9120 from the CLI daemon to the app can leave the app stopped with the port already free: the 1s/2s/4s retry budget (7s) expires while the port is still held. Pre-existing, unrelated to this branch, left for a separate change.

`git revert ac610d4b` returns the isolation alone; the rest survives.


puritysb:
Rebased onto `b087cea6` and re-verified against the parallel work that landed since (App Store screenshot generation, design-system, and `ddefa4f5` which restores `DEBUG` in macOS Debug builds).

**Conflict analysis** — no semantic overlap. Their `apple/` Swift touches `AgentStateHolder` / `AttentionTheaterHUD` / `MonitorScreen`, all UI; this branch is entirely `Daemon/` plus tests and docs. Only `project.pbxproj` overlapped and it is generated (`xcodegen`); the rebase applied cleanly with no conflicts at all.

**The `DEBUG` interaction was the one real risk** — this branch was built and tested without that flag. `Daemon/` and every file changed here contain zero `#if DEBUG` blocks, so it activates no previously-uncompiled daemon code, and the combination now builds and passes with the flag on.

Re-verified on the rebased tree:

- Build succeeds with `SWIFT_ACTIVE_COMPILATION_CONDITIONS: AGENTDECK_APP_STORE DEBUG`
- 453 tests, 1 skipped, 0 failures
- Daemon runs: `/health` `/status` `/devices` all 200 in ~1ms, 7 serial devices enumerated, hooks flowing, WS clients connected
- Zero isolation aborts in the daemon log
- Clean shutdown in 1s; the `Shutdown exceeded 3s — forcing exit` canary did not fire

**System wake is now verified** (it was listed as unverified when this PR was opened). A real suspend/resume produced the full recovery path: `System wake — recovering sessions and devices` → Bonjour re-published → all modules recovered (idotmatrix / timebox / serial / adb / pixoo) → Pixoo PicID cache cleared → 4 stale ESP32 serial connections closed and reopened → Timebox reconnected.

**Known issues, unchanged:**

1. One unexplained crash — SIGBUS in SwiftUI's `RotatingTimelineIcon` / `TimelineView(.animation)` render pass, seconds after the first forced suspend. First occurrence across 28 crash reports, but **no frame from this branch appears in the stack**, so causality is unestablished. It did not reproduce on a second suspend, and the forced `pmset sleepnow` had left the machine in a bad state. Collect the stack first if it recurs.
2. Handing port 9120 from the CLI daemon to the app can leave the app stopped with the port already free — the 1s/2s/4s retry budget expires before the kernel releases TIME_WAIT. Pre-existing, unrelated to this branch.
3. Still unverified: ESP32 connect-burst reentrancy, hook turn serialization, voice-assistant callbacks crossing MainActor↔DaemonActor.

PR #75, opened 2026-07-19
Warn when a serial frame exceeds the board's line buffer

puritysb:
## Why

An ips_10 was reported stuck on "Searching for AgentDeck..." with a live USB cable and a daemon that believed the link was healthy. One hypothesis was an oversized frame. Answering it required adding instrumentation, rebuilding the daemon, and redeploying — for a question that recurs.

The hypothesis was wrong (measured: zero violations), but the check is worth keeping, because the failure mode it guards is silent and easy to misread.

## What oversizing actually does

The firmware does not drop an oversized frame. On overflow it resets `serialBufPos` mid-line, so the **tail** of the frame is parsed as a fresh line, fails the leading `{` test, and never refreshes `lastSerialJsonMs`. Thirty seconds of that and `serialConnected()` flips false — while the daemon keeps writing successfully and sees heartbeat acks. The board shows a searching overlay with a healthy cable attached.

That is exactly the shape of the report that prompted this, which is why it was a plausible suspect.

## The change

One check at `sendToConnection`, the single write chokepoint. Limit mirrors `SERIAL_BUF_SIZE` in `esp32/src/net/serial_client.cpp` (4096; InkDeck 8192). Logs the event type so the offending payload is identifiable without a serial capture, throttled to once per 30s per port+type.

Two existing comments already assumed callers respected this budget — `DaemonServer`'s connect-burst note and `timelineHistoryByteBudget`'s own doc. Now a violation says so instead of being assumed away.

## Verification

- 461 tests pass
- Ran against the live rig before committing: **zero violations** across all four connected serial boards (ips_10, ttgo_t_display, ulanzi_tc001, inkdeck) over a full daemon lifecycle including startup burst

So this ships as a guard, not a fix. The ips_10 report had a different root cause — the board was running firmware from before `dccd1b52` ("avoid IPS10 hosted WiFi teardown race"); flashing current firmware resolved its SDIO assert reboots (36 daemon starts, 0 asserts, versus ~16% per start before).

### DIFF
--- a/apple/AgentDeck/Daemon/Modules/ESP32Serial.swift
+++ b/apple/AgentDeck/Daemon/Modules/ESP32Serial.swift
@@ -818,6 +818,22 @@ actor ESP32Serial {
         let hadLineReset = conn.needsLineReset
         let prefix = conn.needsLineReset ? "\n" : ""
         let payload = Array((prefix + json + "\n").utf8)
+
+        // Invariant: a frame must fit the board's line buffer. Violating it
+        // desyncs the board's parser rather than dropping one frame — see
+        // `lineBufferBytes(for:)`. Log loudly with the event type so the
+        // offending payload is identifiable without a serial capture.
+        let limit = Self.lineBufferBytes(for: conn.deviceInfo?.board)
+        if payload.count > limit {
+            let type = Self.frameType(json) ?? "?"
+            DaemonLogger.shared.throttledDebug(
+                "ESP32",
+                key: "oversized-frame:\(conn.port):\(type)",
+                "OVERSIZED \(conn.port): \(type) frame is \(payload.count)B > \(limit)B line buffer "
+                    + "(board=\(conn.deviceInfo?.board ?? "?")) — board parser will desync",
+                minInterval: 30
+            )
+        }
         switch writePayload(payload, to: conn.fd) {
         case .success:
             conn.needsLineReset = false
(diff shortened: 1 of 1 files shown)


puritysb:
Closing without merge, on re-review against repo conventions.

Two strikes I missed when I proposed keeping it:

1. **It adds a new hand mirror.** `lineBufferBytes(for:)` hand-copies the firmware's `SERIAL_BUF_SIZE` (4096 / InkDeck 8192) into Swift with no drift gate — exactly what CLAUDE.md's SSOT-first rule forbids ("never add a new hand mirror"). If the firmware buffer ever changes, this constant silently lies.
2. **The invariant is already enforced upstream.** Frame size is managed at shaping time — `timelineHistoryByteBudget` (3500B), `limitUtf8Bytes` text caps, session caps. This send-time check is a second line of defense that measured **zero violations** across all four boards over a full daemon lifecycle, i.e. the upstream budgets are doing their job.

The hypothesis this instrumentation was built to test (oversized-frame parser desync explaining the ips_10 "Searching" screen) was refuted by that measurement; the actual ips_10 issue was stale firmware missing dccd1b52, fixed by flashing.

The durable knowledge — the overflow-desync mechanism (firmware resets `serialBufPos` mid-line, tail fails the `{` test, `lastSerialJsonMs` never refreshes, `serialConnected()` flips false while the daemon sees healthy writes) — is preserved in project notes. If a real violation ever shows up, this diff is ~30 minutes to recreate from this PR's history.

PR #81, opened 2026-07-28
fix(apme): disable user diff drivers in judge/artifact git diffs

doug-w:
## Problem

APME runs `git diff` synchronously on the daemon event loop in two places — `collectDiff` (`bridge/src/apme/runner.ts`, judge-prompt input, 4s timeout) and `saveDiffArtifact` (`bridge/src/apme/collector.ts`, 5s timeout). Both honoured user-configured **external diff drivers** and textconv filters.

The motivating real-world setup: my global `~/.gitattributes` maps `*.xlsx diff=excel`, and `diff.excel.command` points at a Java ExcelCompare wrapper. **JVM startup alone takes more than 4000ms**, and git invokes the driver once per changed workbook. So for any session in a repo with xlsx changes:

- `collectDiff` hits its 4s timeout **every time** → the judge gets `(no diff captured)` — and because the spawn is `execSync`, the entire daemon (all decks/dashboards, hook HTTP ingestion, device approval holds) freezes for the full 4 seconds.
- `saveDiffArtifact` then pays the same cost again with its 5s timeout.
- When the judge backend is unavailable, the 30s eval timer re-enqueues the run and re-pays the 4s stall every cycle.

## Fix

Pass `--no-ext-diff --no-textconv` to both spawns. These diffs are machine input for the judge/artifacts, not human-facing output — driver-mapped files degrade to an instant `Binary files differ` line, which is also strictly better judge signal than the empty string the timeout produced. The JVM never launches; the daemon stall disappears.

(`countDiffLines` in `outcome.ts` was checked and left alone — `--shortstat` never invokes external drivers.)

## Test

New regression test in `apme-runner.test.ts` builds a temp repo with a marker external diff driver, first **proves the driver fires** for a plain `git diff` (guards against a vacuous fixture), then asserts the judge prompt contains the raw unified diff and no driver output.

`apme-runner` / `apme-collector` / `apme-category-e2e` suites pass. (The full suite has pre-existing Windows parallel-run flakiness unrelated to this change — same failures reproduce on clean master.)


puritysb (approved):
Approving this — the diagnosis and the fix are both right.

**Verified locally.** Merged this head onto current master and ran the APME suite: 41/41 pass. The load-bearing part of your fixture is the control assertion — `git diff` really does invoke a `diff.<name>.command` driver by default, so the `--no-ext-diff` expectation is testing real behavior rather than restating a default. That is the difference between this test and one that would still pass if the flags were removed.

The reasoning holds for both call sites. `collectDiff` is a synchronous `execSync` on the daemon event loop, so a slow driver stalls every surface *and* still yields `''` when the 4s timeout fires — the worst of both outcomes, and invisible in the eval output.

**One follow-up, non-blocking — same class, uncovered site.** `bridge/src/review-runner.ts:124,127` collects `git diff HEAD --stat` and `git diff HEAD` as LLM review input. It goes through `execFile` with a timeout so it does not block the event loop the way `collectDiff` does, but user diff drivers still execute, their output still lands in the review prompt, and it still eats the `DIFF_BYTE_CAP` budget. The same two flags apply. Take it here or in a follow-up — your call.

**On the red checks:** they were inherited from master, not caused by this PR. Master's tip carried two drifted gates (`sync-hardware-spec-cards` and the `session-utils.ts` SYNC-HASH pin in `D200HLayoutModel.swift`); both are fixed on master now and its own CI is green. I have updated this branch with the new base, so the checks should come back clean.

PR #82, opened 2026-07-28
fix(apme): judge failure backoff + skip darwin-only judge backends off-macOS

doug-w:
## Problem

Two compounding issues in the run-eval loop, found while investigating daemon stalls on a Windows machine (companion PR: #81):

**1. Failed judges retry in an unbounded 30s hot loop.** A failed layer-2 judge call writes no eval rows, so the run still matches `listUnevaluatedRuns` and the daemon's 30-second eval timer re-enqueues it — forever. Each cycle pays a synchronous `git diff` spawn (see #81) plus a doomed backend call, for up to 5 runs per tick, with no backoff and no terminal state.

**2. The default judge config guarantees that loop on Windows/Linux.** The cost-sensitive default is `foundationModels` with MLX fallback — both darwin-only. A non-macOS daemon with no `~/.agentdeck/settings.json` (i.e. a fresh install) feeds every closed code-category run into the failure loop above. Verified live on a Windows 11 daemon.

## Fix

**Backoff (`ApmeRunner`):** per-run failure tracking with exponential backoff — 1m → 2m → 4m → 8m — then the run is parked until daemon restart after 5 attempts. `enqueue()` swallows re-enqueues inside the window; a successful judge clears the state. Deliberately in-memory: a restart is also how a fixed backend comes back, and the cost of forgetting is a single extra retry — not worth a schema migration.

**Platform gate (`judgeBackendSupported` in `settings.ts`):** run/turn/task judge paths skip cleanly (debug-logged, before any diff collection or network attempt) when the configured backend is `foundationModels` or `mlx` on a non-darwin platform. Per the repo's cost-sensitive-defaults contract, no reachable backend is auto-selected in its place — the user opts into `openai` (Ollama / LM Studio / vLLM / OpenRouter) or `api` via settings.json. A test-injected judge override still counts as usable, so existing test fixtures are unaffected.

## Tests

- `judgeBackendSupported`: darwin supports everything; FM/MLX unsupported on win32/linux; `openai`/`api`/`openclaw` portable. (Platform passed explicitly — host-independent.)
- Backoff: failure blocks immediate re-enqueue and retries after expiry; success clears the state and persists evals; 5th failure parks the run permanently.

APME suites (`runner`, `collector`, `category-e2e`, `task-boundary`, `http`) all pass.


puritysb (approved):
Approving this. The backoff is the right shape — in-memory with a documented reason for not persisting it, exponential, and parked rather than looping forever. Verified locally: merged onto current master, 41/41 in the APME suite, and when I stub out the backoff check in `enqueue` two of the three new backoff tests fail — so they are pinning the gate itself, not just the surrounding plumbing.

Two notes, neither blocking, both about the *platform gate* rather than the backoff.

**1. Blocking `mlx` off-darwin looks over-broad.**

`judgeBackendSupported` rejects `mlx` on every non-darwin platform, but that backend is a plain HTTP POST to a configurable endpoint: `llm.mlx.endpoint` in `settings.json`, defaulting to `http://127.0.0.1:8800` (`shared/src/llm-settings.ts:18`, via `mlxChatUrl()`). A Linux or Windows daemon pointed at an Apple-Silicon MLX server elsewhere on the LAN is a legitimate configuration, and this gate disables it silently — only a `debug()` line says why.

`foundationModels` is the one that genuinely cannot work: it spawns a darwin-only helper, so no configuration makes it reachable off-macOS. My suggestion is to narrow the hard gate to `foundationModels` alone and let `mlx` fall through to the backoff you are adding in this same PR — which already handles "backend unreachable" correctly and self-heals on restart when it becomes reachable. If you want to keep an `mlx` fast-path, key it on the endpoint resolving to loopback rather than on `process.platform`.

**2. The platform skip leaves open the loop it is meant to close.**

The skip returns before any judge attempt, so nothing is recorded in `judgeFailures`, and the run still has zero eval rows — it stays in `listUnevaluatedRuns` and the daemon's 30s timer re-enqueues it forever (`daemon-server.ts:3922`). Far cheaper than before, granted: no diff, no network call.

But `runOne` runs layer 1 before reaching the judge gate, and when `deterministic.enabled` is true while producing zero results (`results.length === 0` → no rows inserted, which is the normal case for a project with no configured commands for its language), those deterministic commands re-execute every 30 seconds indefinitely. Parking the run in `judgeFailures` at `+Infinity` on the platform skip — the same terminal state your `JUDGE_RETRY_MAX_ATTEMPTS` path uses — would close that.

**On the red checks:** inherited from master's tip, not from this PR. Both drifted gates are fixed on master and its CI is green; I have updated this branch onto the new base.

PR #83, opened 2026-07-28
fix(docs): re-render hardware spec cards (unbreaks the design-system gate on master)

doug-w:
Master's tip `de9bdb6c` edited the T-Display-S3-Pro **Input** rows in the `docs/hardware-compatibility.md` SSOT table ("3 buttons" → "4 physical buttons (BOOT + split rocker are app-readable; RST is hard reset)") without running `node scripts/sync-hardware-spec-cards.mjs`, so the generated card block in `docs/hardware/index.html` drifted. The design-system workflow's **Verify ESP32 spec-card parity** gate has failed on every run since — on master itself and on every PR that inherits it (e.g. #62).

This PR is exactly the sync script's output, no hand edits: the two T-Display-S3-Pro spec cards pick up the new Input text.

Ran the full "Token sync + lint regression" job steps locally: token mirror sync + self-test, design-system catalog check (27 docs), docs link/anchor check (87 files), and spec-card parity (13 boards in sync) all pass. Two steps aren't measurable on a Windows checkout (the GNB byte-compare and the lint counter both misreport on CRLF working copies — they misreport identically on the last green master commit) — this PR's CI run is the authoritative check for those, and the diff touches only generated card text either way.


doug-w:
Second commit: CI on this PR exposed a **second** pre-existing master breakage hiding behind the spec-card failure — ci.yml's "Preview mirror sync check" has also been red since `8fd0c7d2` (2026-07-26), which added `rawSessionId`/`sameSession` to `shared/src/session-utils.ts` without refreshing the `D200HLayoutModel.swift` SYNC-HASH pin. Those are id-normalization helpers only — no sort/fold/label logic the D200H hand-port mirrors — so per the checker's protocol this is a confirmed-no-impact pin refresh (`91cf2510` → `ae936f86`), no re-port needed. `check-preview-mirror-sync` passes locally (10 pins, 6 mirrors). With both commits this PR should return master to fully green.

puritysb:
Your diagnosis is exactly right, and the write-up saved me the investigation: master's tip did edit the T-Display-S3-Pro **Input** rows in `docs/hardware-compatibility.md` without re-running the sync script, and the generated card block in `docs/hardware/index.html` drifted as a result. It had been failing `Verify ESP32 spec-card parity` on every run since, on master and on every PR inheriting it. Thank you for catching it and for keeping the change to pure generator output.

I am closing this as superseded rather than merging, and the reason is worth stating plainly: the fix already existed locally in unpushed work on my side, including the second half you added — the `session-utils.ts` SYNC-HASH pin refresh in `D200HLayoutModel.swift`, where my local commit lands on the identical hash `ae936f86…`. Those commits are now pushed, so master carries both fixes and its CI is green on `CI`, `Design System`, `Apple Tests` and `Pages` as of `5a60bc89`. Merging this on top would only re-apply the same bytes.

That is a gap on my end, not yours — you could not have known the fix was sitting unpushed, and the correct read of the evidence available to you was the one you made. The general lesson is mine to absorb: an SSOT edit and its generated output belong in the same commit, which is precisely what the gate exists to enforce.

Your other two PRs (#81, #82) are approved, and #62 is in good shape — details in those threads.

PR #86, opened 2026-07-29
Surface per-model weekly usage limits (Fable) on the Pixoo HUD

bruna-peteco:
## Problem

The OAuth usage payload (`api.anthropic.com/api/oauth/usage`) carries a `limits[]` array alongside `five_hour` / `seven_day`. Its `weekly_scoped` rows describe **per-model carve-outs** of the weekly pool. On a Max plan today that means Fable, which draws from its own slice with its own reset:

```json
{ "kind": "weekly_scoped", "group": "weekly", "percent": 12, "severity": "normal",
  "resets_at": "2026-07-29T15:59:59Z",
  "scope": { "model": { "id": null, "display_name": "Fable" } }, "is_active": true }
```

`usage-api.ts` reads only the two top-level windows and discards the array, so `UsageEvent` has nowhere to carry this and the HUD cannot show it. A user burning through their Fable allowance sees `seven_day` creeping and no explanation.

## Change

**Data path.** `parseScopedLimits()` picks the `weekly_scoped` rows (skipping `session` / `weekly_all`, which duplicate the fields already read) and they flow through `ApiUsageData.scopedLimits` → `UsageEvent.scopedLimits`. Rows are keyed by `scope.model.display_name` because `scope.model.id` is commonly `null`. They ride the existing `subscriptionQuotaApplies` gate — a per-model carve-out of a plan quota is meaningless on API billing.

Caches written before this change have no such key, so `readFileCache()` normalizes it to `[]` on read rather than making every consumer guard.

**Pixoo HUD.** Each scoped limit gets its own 7px band under Claude's. Bands now stack upward from the bottom edge (`64 - n * 7`) instead of choosing between two hardcoded offsets, so a third row lands without overflowing. A scoped row is a sub-limit rather than a provider, so it docks the model's initial where a provider draws its official mark; the 3×5 font gains `C F O S` — the initials the payload can scope today (`seven_day_opus`, `seven_day_sonnet`, `seven_day_cowork` are already reserved keys) — plus a solid-chip fallback via `fontCanRender()` for anything else.

**Session-state tally.** Separately, the top-edge strip now counts live sessions by state (processing / awaiting / idle) instead of drawing one dot per agent type. Creature shape already carries agent type; "how many are waiting on me" had no readout at all, and that is the thing you most want at a glance across a tank of six.

## Verification

Ran against a live Max account with an active Fable sub-limit, on real hardware (Pixoo64, firmware reporting `Hardware: 92`):

- HUD renders `9% 3h59` / `8% 6h55` for Claude and a second band `F  12% 6h55` for Fable, matching the payload.
- Tally reads `•1` processing / `•8` idle against the daemon's live session list.
- `pnpm build` clean; `pnpm generate-protocol` re-run and the regenerated Swift/Kotlin/JSON-schema bindings are included, so `protocol-generated-sync` stays green.
- Full suite: 2042 passing across 120 files (3 new cases in `usage-event.test.ts` covering forwarding, absence, and the API-billing gate).
- `eslint` clean on every touched file — no new warnings.

## Notes

- Purely additive on the wire. Clients that ignore `scopedLimits` are unaffected.
- Only the Node renderer is updated. `apple/AgentDeck/Daemon/Modules/PixooRenderer.swift` mirrors this file and would need the same treatment to keep the two surfaces identical — happy to follow up there if you'd like it in the same PR.
- The tally replacing the agent-type dots is a judgment call about what is worth the top seven rows. If you'd rather keep the dots, that part is easy to drop without touching the scoped-limit work.

### DIFF
--- a/bridge/src/__tests__/usage-event.test.ts
+++ b/bridge/src/__tests__/usage-event.test.ts
@@ -44,10 +44,18 @@ function usage(overrides: Partial<ApiUsageData> = {}): ApiUsageData {
     extraUsageMonthlyLimit: 100,
     extraUsageUsedCredits: 12,
     extraUsageUtilization: 12,
+    scopedLimits: [],
     inferredBillingType: 'subscription',
     ...overrides,
   };
 }
+
+const FABLE_LIMIT = {
+  modelName: 'Fable',
+  percent: 11,
+  resetsAt: '2026-06-12T18:00:00Z',
+  severity: 'normal',
+};
 
 describe('buildUsageEvent subscription quota scoping', () => {
   it('omits Anthropic quota when the model is not yet known', () => {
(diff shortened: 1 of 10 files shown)


bruna-peteco:
Closing — opened in error, not something I intend to pursue. Sorry for the noise.

PR #93, opened 2026-08-01
feat(cli): env-var default args for session commands

doug-w:
## What

Two layers of environment-variable defaults so repeated flags can be set once in a shell profile:

- **`AGENTDECK_COMMANDER_ARGS`** — args for the `agentdeck` (commander) layer. Applied only when the top-level command (`argv[2]`) is a session command (`claude`/`codex`/`opencode`/`monitor`) — every other command ignores it, even when a positional value happens to equal a session word (`agentdeck speak board1 claude`). Tokens are spliced *before* any typed flags: **scalar** options are overridden by retyping them (last-write); **boolean** flags have no inverse spelling, so the per-invocation override is **`--no-env-args`**, which disables both env layers for that invocation.
- **`AGENTDECK_CLAUDE_ARGS` / `AGENTDECK_CODEX_ARGS` / `AGENTDECK_OPENCODE_ARGS`** — args appended to the spawned agent command, weaving onto an explicit `-c` instead of replacing it (`-c "claude --resume X"` + `AGENTDECK_CLAUDE_ARGS="--remote-control"` → `claude --resume X --remote-control`).

## How

- `tokenizeArgString` — pure-JS quote-aware tokenizer (no shell, no expansion), so the commander-layer var behaves identically on Linux/macOS/Windows.
- `applyGlobalEnvArgs` — argv splice in `runCli` before `program.parse`, keyed on `argv[2]`; skips entirely when the user typed `--no-env-args`, and strips an env-smuggled `--no-env-args` so the hatch is typed-only (never a half-disabled state).
- `resolveAgentCommand` / `weaveAgentCommand` — the session actions weave the per-agent var onto the `-c` command unless `--no-env-args` was typed; the PTY re-parses the result via the platform shell path that already handles `-c` today.
- `--no-env-args` is registered on all four session commands. It covers only the two `*_ARGS` layers — `AGENTDECK_REMOTE_DAEMON`/`AGENTDECK_DAEMON_HOST` remain separate remote-attach opt-ins.

## Docs

`docs/cli.md` (§ Sessions — env defaults, scalar/boolean override contract, `--no-env-args`), `docs/windows.md` (cmd.exe single-quote caveat), `CLAUDE.md` CLI section + Documentation Index row, devlog entries (2026-07-29 feature, 2026-08-01 review fixes).

## Tests

30 unit tests (`bridge/src/__tests__/cli.test.ts`): tokenization, splice position, argv[2] keying (incl. the `speak board1 claude` non-session regression), hatch skip + env-smuggle strip, per-agent weave — plus the review-requested parse-level coverage: `parseOptions` against the real registered `claude` subcommand (env boolean sticks / `--no-env-args` overrides / scalar last-write through `parseWeight`) and a `parseAsync` end-to-end with a mocked `startSession` proving the action wiring (woven vs unwoven command, env flag presence/absence). `pnpm docs:check` green.

## Review changes (2026-08-01)

1. Session-command detection keyed to `argv[2]` (was: token scan that spliced into `agentdeck speak board1 claude`); regression added.
2. `--no-env-args` per-invocation escape hatch covering both env layers, since boolean flags have no inverse spelling; docs corrected from "explicit flag always wins" to the scalar/boolean split; tested through actual option parsing and action wiring.
3. Rebased onto master `8c65f438` (post-#53); remote-attach option blocks and `warnIfDaemonHostIgnored`/`deriveRemoteAttachOpts` derivation preserved; tokenizer examples corrected to real syntax (`--remote-daemon --daemon-host u2.lan`).


puritysb (changes requested):
The direction is useful, the branch is clean, and both checks are green, but two behavior gaps need to be fixed before these can be called environment-variable **defaults**.

1. **Non-session commands are not actually a no-op when a positional value happens to equal a session command.** `applyGlobalEnvArgs()` scans every argv token from index 2 and injects after the first `claude`/`codex`/`opencode`/`monitor` it sees. For example, `agentdeck speak <board> claude` or another non-session command with that positional value gets `AGENTDECK_COMMANDER_ARGS` spliced into it. The current no-op test only covers a non-session argv with no session-shaped value. This CLI has a fixed top-level command at `argv[2]`; key the decision to that token (or parsed command identity), and add a regression for a non-session positional value equal to `claude`.

2. **One-way boolean env flags cannot be overridden by an explicit CLI invocation.** Inserting env tokens before typed tokens gives last-write semantics for scalar options such as `--weight` or `--daemon-host`, but not for the flags this feature is most likely to persist: `--local`, `--no-postit`, `--no-adb`, hook-skip flags, and `--remote-daemon` after #53. Most have no inverse spelling, so once a shell profile supplies one, the user cannot make an explicit CLI flag “win”; they must mutate/unset the environment for that process. The docs currently promise the stronger default/override contract. Please add a clear per-invocation escape hatch (ideally covering both commander and per-agent env layers), or add complete inverse semantics, and test it through actual option parsing rather than only asserting the transformed argv array.

Sequencing note: #53 is now approved and edits the same CLI option blocks. Please rebase #93 after #53 lands and preserve its remote-attach flags/warning derivation. The tokenizer test should also use the real syntax `--remote-daemon --daemon-host u2.lan`; `--remote-daemon u2.lan` is not a valid invocation under the #24/#53 contract.

puritysb:
@doug-w #53 has now landed on `master` as `8c65f438`. Please rebase this branch onto current `master` while addressing the two requested changes, and preserve the newly merged remote-daemon option/warning derivation. I’ll re-review the updated head when it’s ready.

doug-w:
Both requested changes are addressed at head `08a6c087`, rebased onto `8c65f438` (post-#53). Point by point:

### 1. Session-command detection keyed to `argv[2]`

`applyGlobalEnvArgs` no longer scans argv for a session word — it reads exactly the fixed top-level command position (`argv[2]`) and returns the input untouched for everything else. `agentdeck speak board1 claude` is now a true no-op, locked by a regression with that exact shape. (I checked the structural premise before keying on the token: `program` declares no options that could legally precede the subcommand, no `enablePositionalOptions`/aliases, and the bin entry always passes `process.argv` — so `argv[2]` is the command identity for this CLI.)

### 2. `--no-env-args` per-invocation escape hatch, both layers, tested through parsing

Chose the escape hatch over inverse spellings (a per-flag inverse matrix would keep growing and still not cover the per-agent layer). `--no-env-args` is registered on all four session commands:

- **Commander layer**: the splice is skipped entirely when the typed argv contains the flag — checked pre-parse on the raw argv, because post-parse the env tokens would already have parsed into opts. A `--no-env-args` smuggled inside `AGENTDECK_COMMANDER_ARGS` itself is stripped, so the hatch is typed-only and can never produce a half-disabled state.
- **Per-agent layer**: the actions route through `resolveAgentCommand(agent, opts.command, opts.envArgs !== false)` — the weave is bypassed for the invocation.
- **Scope**: the hatch covers the two `*_ARGS` layers only; `AGENTDECK_REMOTE_DAEMON`/`AGENTDECK_DAEMON_HOST` remain the separate remote-attach opt-ins, and the docs now say so explicitly.

Tests go through actual option parsing as requested, at two levels: `parseOptions` against the real registered `claude` subcommand proves an env boolean sticks as the default, `--no-env-args` genuinely overrides it (`opts.local` undefined, `opts.envArgs === false`), and a retyped scalar wins last-write through the real `parseWeight` processor (`--weight 5` env vs `--weight 2` typed ⇒ parsed number `2`); and a `parseAsync` end-to-end with a mocked `startSession` proves the action wiring itself — woven vs unwoven spawn command and env-flag presence/absence in the options `startSession` receives (this level is what would catch an inverted hatch predicate).

Docs corrected everywhere the old contract was stated: `docs/cli.md` now documents the scalar/boolean split (retype to override a scalar; `--no-env-args` for booleans) instead of "an explicit flag always wins", same correction in `CLAUDE.md` and the `applyGlobalEnvArgs` JSDoc, and the PR body is updated. The devlog records the correction as a new dated entry (the original entry stands as written).

### Sequencing

Rebased onto `8c65f438`; the #53 remote-attach option blocks and the `warnIfDaemonHostIgnored`/`deriveRemoteAttachOpts` derivation are untouched on all four session commands. The tokenizer tests now use the real syntax — `--remote-daemon --daemon-host u2.lan` — in all three places the old `--remote-daemon <value>` shape appeared (including the non-session no-op test).

Suite: 30 cli tests green, full run green apart from the known Windows-host-only failures in files this branch doesn't touch, `pnpm docs:check` green.

puritysb (changes requested):
The two original blockers are addressed at `08a6c087`: command detection is correctly keyed to `argv[2]`, and the typed `--no-env-args` hatch disables both env-default layers with real parser/action coverage. The #53 CLI option blocks and warning derivation are preserved.

One contract wording issue remains before approval. `docs/cli.md` currently calls `AGENTDECK_REMOTE_DAEMON=1` and `AGENTDECK_DAEMON_HOST` “separate remote-attach opt-ins,” and the PR body uses the same plural framing. Under the #24/#53 contract, only `AGENTDECK_REMOTE_DAEMON=1` is the opt-in; `AGENTDECK_DAEMON_HOST` is only an endpoint hint and is inert without that opt-in. Please make that distinction explicit everywhere this PR added or updated the wording, so we do not reintroduce the ambiguity #53 closed.

Also, #95 has now landed on `master` as `64d196ea`. Please rebase onto current `master`, preserve both PRs’ `docs/windows.md` additions, and rerun the suite. Since #95 removes the previously noted Windows-host-only failures, the rebased #93 should now be able to report a fully green Windows run rather than carrying that exception.

puritysb (approved):
Re-reviewed at `4ee551eb`. The remaining remote-attach wording is now correct: `AGENTDECK_REMOTE_DAEMON=1` is the opt-in, while `AGENTDECK_DAEMON_HOST` is only an inert endpoint hint without it. The branch is rebased onto #95, preserves the Windows documentation from both changes, and both checks are green. Approved.

puritysb (approved):
Re-reviewed after merging current `master` into the contributor branch at `f1b75ee7`. The only conflict was the ordering of the #93 and #96 development-log entries; both records are preserved. The CLI changes remain unchanged, the 30 focused CLI tests and documentation check pass locally, and both GitHub checks pass on the updated head. Approved for merge.

PR #95, opened 2026-08-01
fix(tests): green the Windows suite — LF working tree + platform-aware tests

doug-w:
## Summary

`pnpm test` on Windows failed with 13 tests across 7 suites — none tied to any feature work, all environmental. This PR makes the full suite and every locally-runnable PR check green on Windows without weakening any drift gate and without changing ubuntu CI behavior.

- **`.gitattributes` (new)** — `* text=auto eol=lf` + `*.bat text eol=crlf`. The index is already 100% LF (verified: `git add --renormalize .` stages nothing), so this only governs working-tree checkout. CRLF smudging under `core.autocrlf=true` broke bash scripts (`set -o pipefail\r`), vitest imports of the `.mjs` generators, and the golden byte comparisons in the terrarium / session-weight drift gates.
- **`version-policy.test.ts`** — `fileURLToPath` instead of `URL.pathname` (`'/E:/…'` → `resolve()` → `'E:\E:\…'` ENOENT).
- **`linux-service.test.ts`** — the unit-builder tests inherited the harness's Windows tmpdir via `AGENTDECK_DATA_DIR`, which the systemd validator rightly rejects; they now pass an explicit POSIX fixture (no validation bypass, same golden bytes). The one fs-touching test is `skipIf(win32)` — no path is both fs-creatable and `/`-absolute on Windows; ubuntu CI still runs it.
- **`python-ble-runtime.test.ts`** — pins `platform: 'linux'` for the venv-path shape, adds an explicit win32-shape assertion, and normalizes separators in the missing-assets message check.
- **`install.test.ts`** — migration-5 assertion is platform-branched: win32 emits the fire-and-forget PowerShell hook form (no request-response Stop variant exists there); POSIX assertions unchanged. Title updated to state the property honestly.
- **`generate-protocol.sh` + `protocol-generated-sync.test.ts`** — real bug: where `bash` resolves to WSL, custom Windows env vars are dropped entirely, so the drift gate silently regenerated into the repo working tree and compared against an empty temp dir (it never actually gated on Windows). Fixed with `WSLENV=…/p` (WSL propagates + path-translates; Git Bash ignores it and gets the var natively) plus backslash normalization scoped to the override in the script.
- **`docs/windows.md`** — line-endings note with the conversion recipe for pre-`.gitattributes` clones.

## Verification (local, Windows 11)

- Full vitest: **137 files / 2247 passed / 3 skipped / 0 failed** (was 13 failures, confirmed on master first)
- Coverage 48.4 / 44.1 / 47.6 / 49.7 vs thresholds 17/15/14/16
- verify-version, build, typecheck, check-preview-mirror-sync (10 pins in sync), `generate-protocol` → zero diff in the CI gate scope, verify-tokens-sync (+ self-test), design-system `--check`, docs:check, sync-pages-nav `--check`, sync-hardware-spec-cards `--check`: all pass
- Design lint: local totals are grep-version artifacts (same tree bytes: 89 on ubuntu, 91 under WSL grep, 316 under git-bash grep 3.0) — no lint-scoped file is touched by this diff, and CRLF→LF was verified count-neutral

## Ubuntu CI impact

None on any traced path: the env var is unset in CI so the new script block is a no-op; `WSLENV` is inert on Linux; the linux-service and installUnit tests run unskipped; a fresh Linux clone checks out `gradlew.bat` CRLF, which no workflow invokes (all use `./gradlew`).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+# The repo is LF-everywhere, including Windows working trees: bash scripts,
+# vitest-imported .mjs generators, and byte-compared generated mirrors all
+# break under CRLF smudging (core.autocrlf). Only cmd.exe scripts stay CRLF.
+* text=auto eol=lf
+*.bat text eol=crlf
--- a/bridge/src/__tests__/linux-service.test.ts
+++ b/bridge/src/__tests__/linux-service.test.ts
@@ -21,12 +21,18 @@ import {
   installUnit,
 } from '../linux-service.js';
 
+// The vitest harness pins AGENTDECK_DATA_DIR to a host tmpdir (vitest.setup.ts),
+// which on Windows is a `C:\…` path the systemd validator rightly rejects. Tests
+// that don't care about the data dir pass this POSIX fixture instead of
+// inheriting the env, so the unit builder stays pure and cross-platform.
+const posixDataDir = '/tmp/agentdeck-test-data';
+
 describe('buildUnitFile', () => {
   const node = '/usr/bin/node';
   const cliJs = '/home/alice/.local/share/agentdeck/cli.js';
 
   it('produces a well-formed unit with the three standard sections', () => {
-    const unit = buildUnitFile({ node, cliJs });
+    const unit = buildUnitFile({ node, cliJs, dataDirOverride: posixDataDir });
     expect(unit).toContain('[Unit]');
     expect(unit).toContain('[Service]');
     expect(unit).toContain('[Install]');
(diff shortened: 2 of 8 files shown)


puritysb (approved):
Reviewed the Windows-specific failure modes and the Linux/macOS-preserving paths at `eba42929`. The LF checkout policy keeps the canonical index normalized while preserving `.bat` CRLF; the path-shape test fixes retain dedicated platform assertions; and the WSL/Git Bash protocol-output override restores the drift gate instead of weakening it. All five GitHub checks are green, including macOS/iOS. Approved.

Issue #88, opened 2026-07-30
AgentDeck v1.0.2 on Elgato Marketplace marked as Mac Only


doug-w:
Unable to install AgentDeck from Elgato store and previous locally deployed version disabled as manifest is now marked Mac Only: https://marketplace.elgato.com/product/agentdeck-dce3806b-176e-40f2-be7d-e029bec0f464

<img width="705" height="672" alt="Image" src="https://github.com/user-attachments/assets/cbea6296-644b-4e68-908f-700dcd1373bd" />

puritysb:
@doug-w Thanks for reporting this. This is not only a Marketplace-labeling problem; it is a real distribution regression. The published 1.0.2 manifest declares only `OS.mac`, which blocks installation on Windows and disables an existing locally linked plugin. The repository's canonical support documentation explicitly includes Windows 11 for the bridge, hooks, and Stream Deck plugin, so the manifest decision is incorrect.

Rather than only restoring the Windows OS entry, I am also addressing the two dead dials that motivated its removal:

- restore `windows` in the manifest;
- Volume: use Windows PowerShell media keys for rotate/mute behavior;
- Launcher and OFFLINE help: resolve Start-menu apps with a browser fallback;
- pass user-provided targets through environment variables rather than interpolating them into PowerShell source;
- preserve macOS volume readback and the existing browser-tab focus behavior.

I have bumped the Stream Deck channel to **1.0.3**, and the package now passes the official Elgato validator and packaging checks. The 24 focused tests and the full 2,115-test suite also pass. I will link the fix PR and the 1.0.3 submission here when they are ready, and I will ask you to verify the real Windows Marketplace installation plus the Volume and Launcher behavior.

puritysb:
@doug-w The fix has now been merged in #90. Before we resubmit to the Elgato Marketplace, could you please verify this Windows build on your setup?

Download: https://github.com/puritysb/AgentDeck/releases/tag/verification-streamdeck-v1.0.3-windows.1

Please download and double-click `bound.serendipity.agentdeck.streamDeckPlugin` from that pre-release. It is the exact package built from the merged tree; SHA-256 is `53249839ddfc29b4f207d4c81ce7bae93c528c72f51e2f9146ab082889f1b3f7`.

Could you check:
1. Stream Deck accepts and enables the plugin on Windows.
[...]

PR #96, opened 2026-08-01
feat(plugin): cross-platform system layer — Windows joins macOS

doug-w:
@
## Summary

Refs #88. The Stream Deck plugin`s host-control system layer is now cross-platform, so the plugin no longer has to ship **Mac Only** on the Elgato Marketplace.

Host controls (volume, app/URL launch) now dispatch per-platform through `plugin/src/system/` — a single facade over `darwin`/`win32` backends:

- **macOS** — unchanged: `osascript` / `open`.
- **Windows** — a persistent PowerShell CoreAudio coprocess drives master volume (`IAudioEndpointVolume`); `start` / `rundll32` handle URL/app launch. If PowerShell `Add-Type` is blocked (Constrained Language Mode / WDAC) the volume dial shows an N/A face instead of throwing.

The old monolithic `utility-modes/system-control.ts` is replaced by the `system/` facade (`index.ts` dispatch, `types.ts`, `darwin.ts`, `win32.ts`, `win32-volume.ts`). The four encoder actions and the session-slot button dispatch through it.

## Testing

- New/updated unit tests: `system-dispatch`, `system-win32-open`, `system-win32-volume` (replacing the old `system-control` test).
- **Hardware-verified on Windows 11**: confirmed **E1 Volume** dial adjusts the Windows master volume and **E4 Launcher** opens apps/URLs — both working on a physical Stream Deck+.

@


puritysb (changes requested):
Thanks for the follow-up. The persistent CoreAudio helper and the physical Windows verification are valuable, but there is one blocking regression in the current diff.

This branch removes the Stream Deck host push-to-talk path added on `master` in `eb7df238`:

- `session-slot-button.ts` loses the `onKeyUp` end/cancel dispatch and the profile-disappear cancel safeguard;
- `plugin.ts` loses the `voice-ptt-begin` / `voice-ptt-end` / `voice-ptt-cancel` messages and the `voice_state` UI updates.

I also simulated the three-way merge against current `master`; it merges cleanly but the resulting tree still lacks those handlers, so this is a real silent regression rather than a misleading diff.

Please rebase onto current `master` (including #95), preserve the complete PTT path while moving only the system-control imports, and add focused coverage for the key-down/key-up/cancel wiring so a future system-layer refactor cannot remove it unnoticed.

One launcher contract also needs to be checked during the rebase: the published implementation resolves exact Start-menu app names through `Get-StartApps` and rejects a missing app so the `|url:` fallback runs. The new `cmd start` unit test only proves the argv shape because `execFile` is mocked to succeed. Please preserve that resolution/fallback behavior or demonstrate it for the default `Claude` and `Codex` targets plus a missing-app fallback on Windows.

Once those are addressed and the rebased suite is green, I will re-review promptly.

puritysb (approved):
Maintainer follow-up `9953e98e` now addresses the requested changes on this PR branch while preserving the original cross-platform system-layer work:

- merged current `master` / #95;
- restored the complete Stream Deck host PTT lifecycle and `voice_state` UI path;
- extracted key-down/key-up/short-tap/profile-disappear behavior plus daemon wire mapping into focused regression tests;
- kept the safe `rundll32` URL path while restoring exact `Get-StartApps` resolution and missing-app rejection for the Launcher's `|url:` fallback;
- updated the canonical docs and development log.

Verification: 82 focused tests, all 2,274 Vitest tests, full monorepo build, and `pnpm docs:check` pass locally. Both GitHub checks pass on the updated head. Approved.

PR #98, opened 2026-08-02
fix(bridge): resolve Claude usage OAuth credentials on Windows/Linux

doug-w:
## Problem

On Windows, the Stream Deck+ **E2 "Claude Usage"** encoder (`option-dial`) rendered **"No usage data"** even after hours of active Claude Code use.

Root cause: `getOAuthCredentials()` in `bridge/src/usage-api.ts` acquired the OAuth access token **only on macOS**, via the `security` Keychain CLI (`if (process.platform !== 'darwin') return null;`). On Windows/Linux it returned `null`, so `fetchUsageFromApi()` never called the usage endpoint → `cachedApiUsage` stayed `null` → `buildUsageEvent` set `usageStale: true` → the plugin rendered "No usage data". Everything downstream of credential acquisition was already cross-platform.

Claude Code stores OAuth credentials on Windows/Linux in the plaintext file `~/.claude/.credentials.json` (relocatable via `CLAUDE_CONFIG_DIR`) — the same `{ claudeAiOauth: { accessToken, expiresAt } }` JSON the macOS Keychain holds.

## Fix (`bridge/src/usage-api.ts` only)

Refactor credential acquisition into a **pure, injectable resolver** (`resolveOAuthCredentials`), mirroring the existing `judgeBackendSupported(judge, platform)` idiom in `apme/settings.ts`:

- **macOS** — unchanged Keychain read (`security` CLI).
- **Windows/Linux** — `CLAUDE_CODE_OAUTH_TOKEN` env var if set (trimmed), else read `<CLAUDE_CONFIG_DIR ?? ~/.claude>/.credentials.json`.
- A shared `parseOauthPayload()` parses both stores identically (never throws).
- **No subprocess — and no console window — off-darwin**: the `security` CLI is a **lazy thunk** invoked only inside the darwin branch. `buildCredSources()` resolves `CLAUDE_CONFIG_DIR` per call (never hoisted to a module constant) so a post-start relocation is honored.
- `expiresAt` is epoch-ms in both stores, so the existing expiry math is unchanged; env-token path leaves `expiresAt` undefined (existing code attempts the fetch and self-throttles on 401).

`hasOAuthToken()` / `oauthConnected` route through this one function, so they start working on Windows/Linux automatically.

## Tests

Adds `bridge/src/__tests__/usage-api.test.ts` (14 cases) exercising the resolver as a pure function (no `process.platform` mutation): macOS regression, win32 file, env-token precedence + trimming, `CLAUDE_CONFIG_DIR` relocation, missing/malformed JSON → null, and the `buildCredSources` production wiring (asserts the security/file readers are passed as **un-invoked function references** — guarding the "no subprocess on Windows" claim).

## Verification

- `pnpm --filter @agentdeck/bridge build` clean; 62 tests pass across the usage + bridge-core specs (14 new).
- **Live on Windows**: the rebuilt `fetchUsageFromApi()` now returns real subscription usage (5h/7d percentages + reset times, `inferredBillingType: subscription`) where it previously returned `null`. Restarted the daemon and confirmed the E2 encoder pipeline serves live numbers.

## Scope / not included

- No wire-protocol, event-builder, or plugin-renderer changes.
- Windows Credential Manager / DPAPI not used — Claude Code stores a plaintext file on Windows, so a file read is the correct backend.
- Swift/App Store daemon unchanged (it structurally cannot poll quota; has its own empty-state UX).
- An API-key/API-billing account still correctly shows "No usage data" — no subscription windows exist to render.

### DIFF
--- a/bridge/src/__tests__/usage-api.test.ts
+++ b/bridge/src/__tests__/usage-api.test.ts
@@ -1,0 +1,149 @@
+import { homedir } from 'os';
+import { join } from 'path';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { buildCredSources, resolveOAuthCredentials, type CredSources } from '../usage-api.js';
+
+/** Build a CredSources with sane defaults; override per case. Spies let us assert which
+ *  I/O path was taken (critically, that `runSecurityCli` never fires off-darwin). */
+function sources(overrides: Partial<CredSources> = {}): CredSources {
+  return {
+    platform: 'win32',
+    configDir: 'C:\\Users\\test\\.claude',
+    env: {},
+    readCredsFile: vi.fn(() => null),
+    runSecurityCli: vi.fn(() => null),
+    ...overrides,
+  };
+}
+
+const OAUTH_JSON = (accessToken: string, expiresAt?: number) =>
+  JSON.stringify({ claudeAiOauth: { accessToken, expiresAt, refreshToken: 'r', scopes: ['x'] } });
+
+describe('resolveOAuthCredentials', () => {
+  it('darwin: parses the Keychain blob and never reads a file', () => {
+    const readCredsFile = vi.fn(() => null);
+    const runSecurityCli = vi.fn(() => OAUTH_JSON('mac-token', 1_900_000_000_000));
+    const creds = resolveOAuthCredentials(sources({ platform: 'darwin', readCredsFile, runSecurityCli }));
+    expect(creds).toEqual({ accessToken: 'mac-token', expiresAt: 1_900_000_000_000 });
+    expect(runSecurityCli).toHaveBeenCalledOnce();
+    expect(readCredsFile).not.toHaveBeenCalled();
+  });
+
+  it('win32: reads <configDir>/.credentials.json and never spawns the security CLI', () => {
+    const readCredsFile = vi.fn(() => OAUTH_JSON('win-token', 1_888_000_000_000));
+    const runSecurityCli = vi.fn(() => 'SHOULD NOT BE CALLED');
+    const creds = resolveOAuthCredentials(sources({
+      platform: 'win32',
+      configDir: 'C:\\Users\\dougw\\.claude',
+      readCredsFile,
+      runSecurityCli,
+    }));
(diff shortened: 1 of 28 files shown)


doug-w:
Added two commits to complete Windows support here:

**1. Usage relay fallback** (`fa1d98c9`) — the daemon relays usage from a sibling session bridge, but a PTY bridge only refreshes usage when it has *direct* WS clients (dashboards/Stream Deck connect to the daemon, not the bridge). So the relay returned null forever and froze the 5h percentage at its startup value once the OAuth fix let Windows fetch at all. It now falls back to the direct API; the shared 120s file cache dedupes.

**2. Windows console-flashing cleanup** (`6d6b5baf`) — on Windows, several daemon `child_process` spawns lacked `windowsHide`, so each popped a visible console window. The `isLocalConnection` auth path (`getLanIp()` → `route print`) flashed on nearly every request; the passive-observer's `ps` scan flashed every 5s; APME's `pnpm -r build` eval flashed per run. This adds `windowsHide: true` to **every Windows-reachable** exec/spawn across the bridge + shared (git, npm, claude, openclaw, adb, esp32 serial, device-module python, voice, schtasks, detached daemon launches, route/ip probes), and skips APME's shell-based Layer-1 eval on win32 (its grandchild consoles can't be hidden). macOS/Linux-gated calls are left untouched. Flashing verified stopped on Windows.

puritysb (commented):
The OAuth resolver matches Claude Code’s documented Windows/Linux credential storage and looks good. The relay fallback also makes sense.

One semantic concern before merge: commit 6d6b5baf disables APME Layer 1 deterministic evaluation entirely on win32, while the current APME and Windows docs still describe deterministic evaluation as available. Please either remove that behavior change from this PR and track it separately, or document and test the Windows limitation explicitly.

Please also add a regression test for the sibling-relay-miss → direct API fallback introduced in fa1d98c9. The existing tests cover credential resolution but not the relay behavior that fixes the frozen usage value.

doug-w:
APME disabling was a missed debug in tracking down the windowsHide: true and wasn't meant to make it there, I'll push both changes in a moment.

puritysb (approved):
Thanks — both review points are addressed. APME Layer 1 remains enabled on Windows, the relay fallback now has focused regression coverage, and all checks are green.

Issue #145, opened 2026-08-06
Security feedback


jcarberator:
Codex stopped me from installing this because it felt too much was exposed without authentication.  It believed that on a wifi network other users could hijack my agents.

Here is the brief:

Subject: AgentDeck 1.0.2 exposes pairing token and session metadata over LAN

I performed a read-only security review of AgentDeck running on a physical Mac and observed the following:

Environment:

- AgentDeck executable: `/opt/homebrew/bin/agentdeck`
- Installed version: `1.0.2`
- macOS LAN interface: `en10`
- AgentDeck port: TCP `9120`

Findings:

1. AgentDeck listens on a wildcard address

`lsof` reported AgentDeck listening on `*:9120`, rather than only `127.0.0.1`. This makes the daemon potentially reachable from other devices on the local network.

2. LAN accessibility was confirmed

A phone connected to the same Wi-Fi successfully reached the daemon using the Mac’s private LAN address. This confirms that the macOS firewall and network configuration did not prevent another LAN device from connecting.

3. `/health` exposes the pairing token without authentication

Requesting:

`http://<LAN-IP>:9120/health`

returned JSON containing the complete `pairingToken` value. No authentication was required. The credential has since appeared in a screenshot, so it must be treated as compromised. The token itself is intentionally omitted from this report.

4. `/status` exposes operational information without authentication

Requesting:

`http://<LAN-IP>:9120/status`

returned detailed information including:

- Active projects and agents
- Session states and identifiers
- Recent task/event text
- Device names and identifiers
- Stream Deck inventory
- Module availability and configuration status
- Process and daemon metadata

This information was available to any tested device on the same LAN without authentication.

Impact:

A device on the same Wi-Fi network could obtain the AgentDeck pairing credential and inspect potentially sensitive operational metadata. Depending on what the pairing token authorizes, this may also enable unauthorized control or interaction with AgentDeck. That control impact was not tested.

Recommended fixes:

- Bind the daemon to `127.0.0.1` by default.
- Require explicit user confirmation before enabling LAN binding.
- Never include pairing credentials in `/health`, `/status`, logs, or diagnostic output.
- Require authentication for all nontrivial LAN-accessible routes.
- Return only minimal health information, such as status and version.
- Redact project names, task text, session identifiers, and device identifiers from unauthenticated responses.
- Rotate existing pairing tokens when upgrading from affected versions.
- Add automated tests verifying that secrets never appear in HTTP responses or CLI status output.
[...]

PR #150, opened 2026-08-07
fix(security): stop leaking pairing credentials over UDP and logs

jcarberator:
## Summary

- remove the pairing-token parameter and field from the UDP 9121 discovery beacon
- keep UDP/mDNS discovery daemon-only so session project names are not broadcast on the LAN
- replace token-bearing startup URLs with an explicit `agentdeck qr` pairing hint
- document the discovery/auth boundary and add regression coverage

## Security impact

The Node daemon in npm 1.0.9 passed `core.authToken` into `advertiseUdpBroadcast()`, which serialized it into JSON sent every two seconds to both `255.255.255.255:9121` and the subnet broadcast address. Any peer on the LAN segment could recover the pairing credential passively, bypassing the default-deny work in #145. The same credential was also embedded in startup log lines through `core.wsUrl`, persisting it in long-lived daemon logs.

The UDP beacon remains available for discovery but now contains only `v`, `ip`, `port`, `project`, and `agent`. Existing provisioned clients retain their stored token; new clients must obtain it through an explicit pairing/provisioning flow.

Follow-up to #145.

## Verification

- `pnpm test` — 169 files passed; 2,713 tests passed, 1 skipped
- `pnpm --filter @agentdeck/bridge build`
- `pnpm --filter @agentdeck/bridge typecheck`
- `pnpm docs:check`
- targeted ESLint — 0 errors (pre-existing warnings only in large startup files)
- `git diff --check`


puritysb:
Merged as 681a143 — thank you for coming back not just with a second report but with a complete, well-tested fix. We traced every hunk against the runtime paths and confirmed all three findings: the beacon token itself, the token-bearing `Pairing URL:` startup log lines via `core.wsUrl`, and the session bridges still beaconing (even under `--local`) because the `broadcast` key was missing from their module config. The `Reflect.apply` sentinel test is a nice touch — it catches the parameter being restored instead of silently adapting to it.

A small follow-up PR is coming on top of this: the beacon will additionally advertise `authRequired: true` (matching the unauthenticated `/health` semantics, as your issue suggested), with an exact field-whitelist regression gate so any future beacon field addition has to answer the "is this a secret?" question explicitly. The `agentdeck token rotate` advisory will ship with the release notes.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #149, opened 2026-08-07
security feedback #2


jcarberator:
Thank you for the amazing turn around on the prior issue.

### v1.0.9 still broadcasts pairing token over UDP

Confirmed in the published `@agentdeck/bridge@1.0.9` artifact and current `master` by tracing the runtime code path.

Despite the HTTP/mDNS fix in #145, UDP discovery still exposes the credential:

1. `daemon-server.ts` enables the broadcast module and provides `core.authToken`.
2. `modules/broadcast-module.ts` calls:

```ts
advertiseUdpBroadcast(ctx.port, ctx.projectName, this.agentType, ctx.authToken)
```

3. `broadcast.ts` adds `payload.token = token`, then broadcasts the JSON to `255.255.255.255` and the subnet broadcast address on UDP port 9121 every two seconds.

Source: [broadcast-module.ts](https://github.com/puritysb/AgentDeck/blob/master/bridge/src/modules/[broadcast-module.ts](https://github.com/puritysb/AgentDeck/blob/master/bridge/src/modules/broadcast-module.ts#L21-L23)#L21-L23), [[broadcast.ts](https://github.com/puritysb/AgentDeck/blob/master/bridge/src/broadcast.ts#L72-L94)](https://github.com/puritysb/AgentDeck/blob/master/bridge/src/broadcast.ts#L72-L94).

Impact: any device on the network segment can passively obtain the pairing token, then authenticate to the protected HTTP/WebSocket interfaces. This effectively nullifies the new LAN authentication boundary.

`AGENTDECK_LOOPBACK_ONLY=1` does not mitigate it: that changes the TCP listener address, while UDP discovery still uses the machine’s LAN IP and remains enabled.

Suggested fix:

- Remove the token argument and `payload.token` from UDP discovery.
- Advertise only non-secret discovery metadata and `authRequired`.
- Add regression tests asserting that every discovery transport—mDNS, UDP, and unauthenticated `/health`—contains no token or secret-shaped fields.
- Advise users to rotate their token after installing the corrected release.

puritysb:
[...]

PR #163, opened 2026-08-08
feat(codex): surface persistent goals as read-only tiles

jcarberator:
## Summary

- Discover non-complete Codex persistent goals from a user-approved ~/.codex directory using read-only SQLite access.
- Represent goals as synthetic observed sessions with lifecycle status, parent project, update time, and recent worker count.
- Keep goal tiles separate from ordinary Codex session folding and place them after their matching parent project.
- Render distinct goal badges, lifecycle labels, worker activity, and wrapped goal names on Stream Deck.
- Extend the shared protocol and generated Swift, Kotlin, and JSON schema mirrors with optional goal metadata.
- Fail closed: missing files, incompatible tables, and unknown future statuses produce no goal rows.

The feature never starts, pauses, updates, or otherwise modifies a Codex goal.

## Verification

- Full workspace suite: 2,762 passed, 1 skipped.
- Native macOS suite: 550 passed, 1 skipped.
- Workspace build and protocol generation passed.


puritysb (changes requested):
There's careful work here — the status allowlist fails closed on unknown future values, SQL is fully parameterised, the `sessionKind == "goal"` guard is mirrored into Swift `isCodexPayload` as well as the TS `isCodexSession`, the generated protocol mirrors were regenerated, and `docs/appstore-feature-matrix.md` was updated in the same commit. That's the house style followed properly.

I still can't merge it. The blocking problems are about what this puts on the wire and what it does to the daemon, not about the goal-reading code itself.

## 1. Blocking: it breaks every already-shipped client

`codexGoalToDict` appends synthetic rows into `sessions_list`. That event is consumed today by software we do not control the version of:

- **iOS/iPadOS 1.0.4** — live on the App Store right now
- **macOS 1.0.4** — currently in App Review
- **Stream Deck plugin 1.0.2** — published on the Elgato Marketplace
- **Ulanzi D200H plugin** — published
- **ESP32 firmware** — flashed devices in the field, some of which can only be updated over WiFi OTA

None of them know `sessionKind`. They will treat `codex-goal:<threadId>` as an ordinary session: it gets sorted into the deck, **consumes a physical Stream Deck key**, inflates session counts, and appears as a tappable row in the shipped iOS app. Pressing it sends `focus_session` with an id that is in `cachedCodexGoals` but not `cachedSessions` — `DaemonServer.swift:3168` doesn't guard that, so the daemon sets global focus to a session that doesn't exist and every device drops into a detail view with no state behind it. (`session_command` *is* guarded at `:3180`, so steering is at least an inert no-op.)

This collides head-on with our own compatibility rule in CLAUDE.md — "numeric `X.Y.Z` versions are mutually compatible exactly when `X.Y` matches". Everything above is `1.0.x`, so every one of those clients believes it is compatible with this daemon, and `pnpm verify-version` will happily agree.

Adding an **optional field** to `SessionInfo` is safely additive; adding a **new kind of row** to `sessions_list` is not, and we have no negotiation mechanism for it. Two ways out, either is fine by me:

- gate emission on a client-advertised capability (clients already send `client_register` — have goal-aware clients opt in, and emit nothing to everyone else), or
- keep goals off `sessions_list` entirely and ship them as their own event, so an unaware client simply ignores an unknown `type`.

This is the same lesson as the `#145`/`#149` fallout, in the other direction: there, removing a field turned out to be a consumer migration. Here, adding a row kind is one.

## 2. Blocking: an `active` goal is `processing` forever

```swift
"state": isActive ? "processing" : "idle",
```

A persistent goal has no turn boundary, so this drives the spinner, the processing animation, and the attention path with no bounded work behind it. That is exactly what got iPhone/iPad 1.0.2 (3901) rejected under Guideline 2.1(a) on 2026-08-04, and CLAUDE.md carries the rule that came out of it: *"A progress indicator binds to bounded work, and the bound is structural."* A goal's lifecycle status is real information, but it needs its own visual axis — a badge, which this PR already renders — not the running state.

There's an irony worth naming: #162 (just merged) exists to un-stick observed Claude sessions that sit in `processing` forever. This would reintroduce the same symptom by construction.

## 3. Blocking: synchronous I/O on `@DaemonActor`

`LocalCodexGoalObserver.collect()` is called from `refreshSessions()`, which is the **10-second** session poll (`DaemonServer.swift:6422`). Every tick it does, inline on the daemon actor:

- two `sqlite3_open_v2` + prepare/step cycles,
- one `threads` query and one `thread_spawn_edges` COUNT **per goal**,
- `Data(contentsOf: session_index.jsonl)` — the whole file, with no cap, then a full UTF-8 split and a `JSONSerialization` call per line.

The rest of the observer code reaches for `readFileHeadAndTail` with `MAX_TAIL_BYTES` precisely to avoid this. We've already had the daemon stall on slow SQLite work, and `@DaemonActor` exists specifically because starving that executor takes the whole hub down with it. This needs to be off-actor (`Task.detached` + a cached result), bounded, and ideally on a much slower cadence than 10s — a persistent goal ledger does not change every 10 seconds.

## 4. Privacy: the `~/.codex` grant gets silently widened

`AppPreferences.chooseCodexDirectory()`'s panel copy was updated, which is the right instinct — but it only affects **new** grants. Every user who already approved that folder for "Codex plan and usage limits" now has that same security-scoped bookmark used to read their goal ledger, all thread titles and working directories from `state_5.sqlite`, and thread names from `session_index.jsonl`. Reusing an existing consent for a new data category needs a re-prompt, and `apple/APP_REVIEW_NOTES.md` needs to describe the widened read scope since its claims ship to the reviewer.

Related, lower severity: `goalObjective` is free-form user text and goes out over plaintext LAN WebSocket to every paired device. `cwd` and `goal` already travel that path so this isn't a new class of exposure, but it's a wider one, and it should be a deliberate choice rather than a side effect.

## 5. It may never produce a row

`sqlite3_open_v2(path, &db, SQLITE_OPEN_READONLY, nil)` against a **live WAL-mode** database needs to create/attach the `-shm` file; read-only opens of a WAL DB that isn't already checkpointed typically fail with `SQLITE_READONLY_CANTINIT`. Failing closed means we don't crash, which is good — but it also means the feature can be permanently and silently empty, and the CI suite can't tell us. Has this been confirmed against a running Codex app, with a goal actually visible on a deck?

`goals_1.sqlite` / `state_5.sqlite` / `thread_goals` / `thread_spawn_edges` are undocumented, version-numbered internal storage. When Codex ships `state_6`, this disappears without a sound. At minimum log once at debug when the files are present but the schema doesn't match, so the failure is diagnosable.

## 6. Parity policy is inverted

Seven fields (`sessionKind`, `goalStatus`, `parentProjectName`, `goalThreadId`, `activeWorkers`, `goalObjective`, `goalUpdatedAtMs`) are added to the shared `SessionInfo`, but the only producer is the Swift daemon — the Node bridge will never emit them. Our standing policy is that Swift follows Node up to the sandbox limit, not that it leads. If goal observation is worth having, the Node bridge should be the reference implementation (it has no sandbox constraint on `~/.codex` at all) and Swift should mirror it.

## 7. Blocking: preview-mirror pins (the red `test` job)

```
apple/AgentDeck/Rendering/SessionSlotRenderer.swift   pinned 15a76503… / actual 55faf655…
apple/AgentDeck/UI/Preview/Devices/D200HLayoutModel.swift  pinned b08adbcc… / actual 077e6181…
```

Both mirrors need the change ported (or a written confirmation that the mirrored surface is unaffected) plus a pin bump, in the same commit.

## Minor

- `(rhs.goalUpdatedAtMs ?? 0) - (lhs.goalUpdatedAtMs ?? 0)` is a subtraction comparator; the repo requires three-way comparators throughout `session-utils.ts`.
- `readGoals`: if `sqlite3_prepare_v2` fails but writes a non-nil statement, the `guard` returns before `defer` is installed and the statement leaks.
- `ISO8601DateFormatter()` is constructed per goal per refresh; hoist it.
- `wrapSessionName` here is byte-identical to the one #164 adds to the same file. One of the two has to go.

## Where I'd like to land

The underlying idea — surfacing long-running Codex goals as first-class context — is a good one, and I don't want to lose it. But it needs to be (a) invisible to clients that didn't ask for it, (b) not a permanent spinner, and (c) off the daemon actor. If you'd like, split the goal *reader* (`LocalCodexGoalObserver` + its tests) into its own PR — with the WAL question settled and the I/O moved off-actor — and we can land that with no wire exposure at all while the transport question gets decided separately.

puritysb:
Status note, not a new review: #164 has landed on `master`. It adds `plugin/src/device-profile.ts` and changes `session-slot-renderer.ts`, `session-slot-button.ts`, `plugin.ts` and the renderer snapshots — all files this PR also touches, and it carries the `wrapSessionName` that is byte-identical to the one here. This will need a rebase, and the duplicate helper should collapse onto the one now in `master` rather than being reintroduced.

The seven points from my review above are unchanged and still blocking — in particular the `sessions_list` compatibility question (#1), which is the one that decides the shape of the whole feature. No need to rebase until that transport question is settled; I didn't want you to discover the conflict without knowing why it appeared.

puritysb:
Heads-up on the red `test` check — **it is not your change.**

It is our `check-preview-mirror-sync` gate, and it is firing because master moved after you branched:

```
mirror : apple/AgentDeck/Rendering/SessionSlotRenderer.swift
origin : shared/src/svg-renderers/session-slot-renderer.ts
mirror : apple/AgentDeck/UI/Preview/Devices/D200HLayoutModel.swift
origin : shared/src/session-utils.ts
```

Both origins changed on master (session-slot animation, and the session `--weight` ordering), and the pins were updated there in the same commits. Your branch carries the old pins, so the gate sees a mismatch. On current master the gate is green — 10 pins across 6 mirrors, all in sync.

A rebase onto master clears it; nothing to change in your PR for this. Sorry for the noise.

puritysb:
Maintainer direction on the transport question: please keep goals out of `sessions_list` and introduce a separate `codex_goals` event. Existing 1.0.x clients will then safely ignore the unknown event instead of treating goals as sessions.

Please split the work in two:

1. A goal-reader PR with bounded, off-actor I/O, live-WAL verification, schema-mismatch diagnostics, and no wire exposure.
2. A transport/rendering PR after that, with the Node daemon as the reference producer, Swift parity, explicit consent migration, and goal lifecycle rendered as a badge rather than `processing`.

No need to rebase the current cross-surface branch first; extracting the reader onto current `master` will avoid the conflicts with #164.

puritysb:
Closing this implementation branch as superseded by the maintainer direction above. The persistent-goal idea remains welcome, but this PR combines reader, transport, and rendering concerns and now conflicts with current `master`. Please use fresh branches for the two agreed slices: first the bounded/off-actor goal reader with no wire exposure, then a separate `codex_goals` transport and rendering PR. There is no need to rebase or continue this branch.

PR #164, opened 2026-08-08
feat(plugin): add device-aware Stream Deck profiles

jcarberator:
## Summary

- Map every DeviceType currently documented by Elgato to a distinct canonical family, including the correct separation of Stream Deck Pedal and Corsair Voyager.
- Share one mapping between device registration and per-action layout detection.
- Show accurate fallback names for Mobile, GKeys, Voyager, SCUF, Neo, Studio, Virtual, and Galleon devices.
- Use larger 20-point, earlier-wrapping labels on original 72-pixel Stream Deck keys while retaining the standard profile on XL, Plus, and other higher-resolution keys.
- Keep the selection automatic and omit temporary device diagnostics.

Device mapping follows the official Elgato SDK device table: https://docs.elgato.com/streamdeck/sdk/guides/devices/

## Verification

- Focused device and renderer tests: 42 passed.
- Full workspace suite: 2,760 passed, 1 skipped.
- Native macOS suite: 549 passed, 1 skipped.
- Workspace build passed.


puritysb (changes requested):
The consolidation is right — `familyForDeviceType` was duplicated in `plugin.ts` and `session-slot-button.ts` with two *different* tables (the button copy was missing Pedal and defaulted unknown devices to `streamdeck`), and that divergence was a latent bug. Extracting one module is the correct fix. Three things before this can merge.

## 1. Blocking: the preview-mirror pin (this is the red `test` job)

```
mirror : apple/AgentDeck/Rendering/SessionSlotRenderer.swift
origin : shared/src/svg-renderers/session-slot-renderer.ts
pinned : 15a765038d23df870db58302442056340cb66c8a
actual : a943e7e09cc29dbaa75b654a5d6f25c6322e40d7
```

`scripts/check-preview-mirror-sync.mjs` (CI: `.github/workflows/ci.yml:50`) pins the Swift Stream Deck preview to the TS renderer it mirrors. Changing the origin requires, in the same commit, either porting the visual change into `SessionSlotRenderer.swift` or confirming it doesn't affect the mirrored surface — and then bumping the `SYNC-HASH` line to the new blob hash.

Here it genuinely does affect it: `lowResolutionKey` changes font size and the name baseline (`y=128` vs `y=120`), and the macOS app's preview is what users check their layout against. Please port `wrapSessionName` + the 20pt profile into the Swift mirror and update the pin.

## 2. `Number(d.type)` regresses the unknown-device path

```ts
family: familyForDeviceType(Number(d.type)),
```

`Number(undefined)` is `NaN` → `default` → fine. But `Number(null)` and `Number('')` are both **`0`**, which now means `'streamdeck'` — and `usesLowResolutionKeyProfile('streamdeck', …)` returns `true` unconditionally, so a device that merely failed to report a type gets the 72px typography on a 144px key. The previous `d.type as number | undefined` had no such hole. Please keep the cast, or guard with `typeof d.type === 'number' ? d.type : undefined`.

## 3. DeviceType values 3, 4, 6, 8–12 need a citation in-tree

The installed `@elgato/streamdeck` doesn't ship the `DeviceType` enum, so I can't verify 3=Mobile / 4=GKeys / 6=Voyager / 8=SCUF / 9=Neo / 10=Studio / 11=Virtual / 12=Galleon100SD against anything local. The values we already shipped (0,1,2,5,7,13) match, and the Pedal-vs-Voyager separation looks right. Could you add the doc URL as a comment on the switch, and note which SDK/schemas version the table was read from? A wrong number here silently mislabels a device family in the daemon topology row and can't be caught by a test.

## Minor

`wrapSessionName` breaks on spaces and camelCase only, so CJK project names (no spaces, no `[A-Z]`) fall through to a hard `maxChars` cut. Acceptable for now, but a width-aware break would be better on the 72px profile where the budget is tightest.

## Conflict warning

#163 adds a byte-identical `wrapSessionName` to the same file (only the `maxChars` default differs), and both PRs touch `session-slot-renderer.ts`, `session-slot-manager.ts` and the renderer snapshots. Whichever lands first, the other needs a rebase. My inclination is to land this one first — it's the smaller and better-scoped of the two.

jcarberator:
Addressed all three review points:

- Ported the low-resolution session-name wrapping, 20 pt name typography, and adjusted tool baseline into the Swift device preview; refreshed the mirror `SYNC-HASH` and added Swift wrapping coverage.
- Replaced `Number(d.type)` coercion with strict integer validation so missing, null, and string values cannot become DeviceType 0; added regression tests.
- Documented the DeviceType 0–13 mapping with the official Elgato device table and verified it against the current SDK/schema.

Verification: plugin build and typecheck; 174 Vitest files / 2,815 tests passed (1 skipped); macOS suite 556 tests passed (1 skipped); preview-sync and surface-mirror checks passed.

puritysb (approved):
All three review points verified against the code, not just the summary.

1. `deviceTypeFromUnknown()` does a strict `typeof === 'number' && Number.isInteger` check, so `null` and `''` can no longer become DeviceType 0. You went further than I asked by splitting unknown devices into their own `'streamdeck-unknown'` family and gating the low-resolution profile on 5×3 geometry — that closes the hole more cleanly than keeping the old cast would have, and the Swift topology surfaces were updated for the new family string in the same commit.
2. `SessionSlotText.wrapSessionName` is ported into the Swift preview with the 20 pt profile and the adjusted tool baseline, and the SYNC-HASH pin moved `15a76503…` → `a943e7e0…`. `check-preview-mirror-sync` reports 10 pins across 6 mirrors all in sync.
3. The DeviceType 0–13 table now carries the Elgato doc URL and the verification provenance (`@elgato/streamdeck 2.1.0` + manifest schema, dated).

I resolved the `DEVELOPMENT_LOG.md` conflict against master myself and pushed the merge to your branch — both entries kept, no functional change. Verified locally afterwards: build clean, 178 test files / 2893 passed, preview-mirror and surface-mirror gates green; CI agrees on all five jobs.

Heads up for #163: it adds a byte-identical `wrapSessionName` to `session-slot-renderer.ts` and touches the same snapshots, so it will need a rebase now that this has landed.

Issue #143, opened 2026-08-06
Windows: map Codex/ChatGPT Desktop processes to rollout sessions


puritysb:
## Context

Ulanzi's review team (D200H marketplace) evaluates AgentDeck on Windows machines that have the **Claude / ChatGPT desktop clients but no CLIs**. On macOS the passive observer already detects Codex sessions launched from the ChatGPT desktop app automatically; on Windows the observer is effectively blind, and we committed to closing that gap.

Three concrete holes in `bridge/src/passive-observer.ts`:

1. **Process scan** shells out to `ps -ww -eo pid=,ppid=,rss=,tty=,command=` — on Windows that only exists under git-bash, and even then it doesn't report native Win32 process command lines. Needs a native backend (`Get-CimInstance Win32_Process` via PowerShell).
2. **pid → rollout mapping** uses `lsof` (macOS) / `/proc/<pid>/fd` (Linux). Neither exists on Windows, so Codex sessions are dropped even if the scan found the process.
3. **Host-app resolution** (`appNameFromCommand`) matches macOS `/*.app/` bundle paths only — a tty-less Codex under `ChatGPT.exe` would never resolve its owning app.

## What we know from the macOS measurement (2026-08-04, live)

- ChatGPT.app runs `…/ChatGPT.app/Contents/Resources/codex -c … app-server`; **one app-server pid holds one open rollout file per open conversation (1:N)**.
- The desktop app **does fire the `codex_*` lifecycle hooks** from `~/.codex/config.toml`, carrying `session_id` (= rollout id) + `cwd` — so hooks are a viable identity source where file-handle enumeration isn't.
- OTel spans carry no conversation identity at all (tokio thread ids only) — not usable.

All of these are **unmeasured on Windows**. Whether the Windows ChatGPT client even bundles Codex the same way, whether it reads `%USERPROFILE%\.codex\config.toml`, and what its process tree looks like are open questions — and the answers decide the design.

## Ask

@doug-w — you built the Windows system layer (#96), greened the Windows suite (#95), and did the Windows OAuth path (#98), so you know this surface better than anyone. **Would you like to take this end-to-end?** We'd review, and happily pair on the observer-side plumbing which is currently macOS-shaped.

If you don't use Codex/ChatGPT on Windows, a diagnostics capture alone would already unblock us — with a Codex conversation open in the ChatGPT desktop app, and separately a terminal `claude` session running:

1. ```powershell
   Get-CimInstance Win32_Process |
[...]

PR #165, opened 2026-08-08
Windows codex micro

Roja430:
バグ修正


puritysb:
@Roja430 — I read through this before you closed it, and I want to make sure the work doesn't get lost along with the PR.

Closing it was the right call in its current shape: 49 commits and ~9.2k lines against `master`, plus some local state that rode along (`"@agentdeck/bridge": "link:"` in `bridge/package.json` and the new `bridge/pnpm-workspace.yaml` override — artifacts of a local `pnpm link`, and the final commit is `一旦コミット`). Nobody can review that as one unit. But the commits underneath are individually well-scoped and clearly titled, which is more than I can say for most large PRs.

Two of them are fixes to *our* tooling, and I've verified both and opened them separately, with credit to you:

- **#166** — `generate-protocol.sh` writes each schema with `> "$OUT_DIR/…"`, which truncates the target before the generator runs. Since `OUT_DIR` defaults to the repo, a failed run empties a committed artifact. Reproduced with a stub generator that exits 1: on `master` `bridge-event-schema.json` drops to 0 bytes. Your `WSLENV` explanation for why the drift gate's `AGENTDECK_PROTOCOL_OUT_DIR` override doesn't reach WSL's bash is exactly right, and it's what turns this from theoretical into routine on Windows.
- **#167** — design lint R6. Your title ("had never matched an emoji") is literally true: the bracket set collapses to one `\x98-\xE2` range, and in a UTF-8 locale grep rejects the whole pattern as an illegal byte sequence — on stderr, which both `scan()` and the CI step discard. The rule reported a clean codebase for its entire life. I added a boundary-case regression test so it can't quietly die again.

Thank you for both — a gate that fails silently is the worst kind, and neither would have been found by anyone not running on Windows.

## If you'd like to land the rest

Please do, but in reviewable slices — one concern per PR, branched off current `master`, with the `link:` lines and `pnpm-workspace.yaml` dropped. Rough order of how easily I think they'd go in:

1. **Windows console flashing** (`fix(windows): stop hooks flashing a console on every tool call`, `fix(bridge,shared): stop console windows flashing on Windows`, `ea6a4313`). This overlaps #143, which is open and ours. Highest value, self-contained.
2. **Small correctness fixes** — `stop launched agents inheriting the parent session identity`, `leave the home directory out of recent projects`, `stop probing adb on every /devices request`, `stop a focus switch from killing the daemon`. Each stands alone.
3. **Plugin/encoder behaviour** — the AWAITING option strip, the offline banner column fix, `stop repainting encoder LCDs that have not changed`. These touch shared renderers, so please check `scripts/check-preview-mirror-sync.mjs` before pushing: changing `shared/src/svg-renderers/*` requires porting the change into the Swift preview mirror and bumping its `SYNC-HASH` pin in the same commit. That gate is currently red on two other open PRs for exactly this reason.
4. **Larger features** — cost tracking + daily budget, session launch/fork from the deck, the model/effort dials. Each is its own design conversation; one PR each, and please open an issue first for the launch/fork ones so we can agree on the shape before you spend the effort. Worth knowing up front: the macOS App Store build can't spawn processes at all, so anything that launches a session has to be CLI-tier only.

One flag on `feat(bridge): add an opt-in verbatim PTY capture`: it's correctly opt-in via `AGENTDECK_PTY_RAW_LOG` and off by default, but it appends raw terminal output — which can include credentials and tokens — to an unbounded, unrotated file. If you resubmit it, it needs a size cap or rotation and a line in the docs saying plainly what lands in that file.

日本語でも大丈夫です。PR の説明は英語だと助かりますが、質問は日本語で構いません。

PR #168, opened 2026-08-09
feat: add read-only Hermes session observation

jcarberator:
## Summary

- install an opt-in AgentDeck observer plugin under `~/.hermes/plugins/agentdeck-observer/` from `agentdeck daemon install`
- translate official Hermes session, prompt, tool, and completion callbacks into display-only observed sessions
- add Hermes to the shared protocol and capability model without terminal, steering, options, usage, or model-catalog capabilities
- avoid bundling or substituting an unverified Hermes brand mark

## Safety and scope

- loopback-only HTTP with an 800 ms timeout; observer failures are swallowed
- no Hermes database reads, remote relay, task control, or response mutation
- no subprocess path or installer UI added to the App Store app
- interrupted processing rows expire after three minutes
- `AGENTDECK_NO_HERMES_HOOKS=1` disables installation

The integration follows Hermes official plugin and lifecycle-hook contracts:
- https://github.com/NousResearch/hermes-agent/blob/main/website/docs/user-guide/features/plugins.md
- https://github.com/NousResearch/hermes-agent/blob/main/website/docs/user-guide/features/hooks.md

## Verification

- `pnpm build`
- `pnpm typecheck`
- `pnpm test` — 2,766 passed, 1 skipped
- `pnpm docs:check`
- `pnpm check-surface-mirrors`
- `pnpm generate-protocol`
- changed-file ESLint — no errors
- generated Python plugin syntax compilation
- contribution-only privacy and secret scan


jcarberator:
Rebased onto the current `master` and resolved the development-log overlap while preserving both entries. No functional Hermes code changed during the rebase.

Verification after rebase: build and typecheck passed; 175 test files / 2,820 tests passed (1 skipped); documentation, surface-mirror, protocol-generation, diff, and privacy checks passed.

puritysb (changes requested):
Thank you for this — and before anything else, a caveat that should change how you spend time on it: **whether AgentDeck takes a Hermes integration at all, and at what tier, is not settled yet.** That is my call to make and I haven't made it (more at the bottom). Please don't push a revision on the strength of this review alone. Everything below is here so that if we do land Hermes, none of the analysis has to be redone — and because two of the findings are worth knowing regardless of what we decide.

First, what I checked and found correct. I verified `NousResearch/hermes-agent` upstream rather than taking the PR description's word for it: the plugin layout (`~/.hermes/plugins/<name>/{plugin.yaml,__init__.py}`), the `register(ctx)` entrypoint and `ctx.register_hook()` are all as described. I read the shipped `plugins/observability/langfuse` plugin, and hooks are dispatched with **keyword** arguments (`def on_pre_tool_call(*, tool_name="", session_id="", ...)`), so your `kwargs.get("session_id" / "model" / "platform" / "tool_name")` reads match the real contract. On our side, the `hookAgentType === 'hermes'` early return is placed correctly — ahead of the timeline, state machine, permission and APME paths — so "display-only" is true in the code and not just in the description. And the 3-minute `PROCESSING_TTL_MS` is the right lesson carried over from #162. That is careful work.

## 1. Blocking: this defeats a consent gate Hermes built on purpose

From upstream's plugin documentation:

> **General plugins and user-installed backends are disabled by default** — discovery finds them (so they show up in `hermes plugins` and `/plugins`), but nothing with hooks or tools loads until you add the plugin's name to `plugins.enabled` in `~/.hermes/config.yaml`. **This stops third-party code from running without your explicit consent.**

`hermes plugins install owner/repo` respects this too — it prompts `Enable 'name' now? [y/N]`, defaulting to **no**.

`hermes-install.ts::enableHermesPlugin()` runs `execFileSync(hermes, ['plugins', 'enable', 'agentdeck-observer'])`. It is called from `agentdeck daemon install` on all three platforms, unconditionally, with an environment variable as the only way out. So AgentDeck writes third-party code into the user's plugin directory and then adds itself to the allow-list that exists specifically to stop that. The PR describes this as opt-in; as written it is opt-out.

Two things make it stickier than a single bad call:

- Upstream grandfathers on upgrade: user plugins already under `~/.hermes/plugins/` and not in `plugins.disabled` are **automatically moved into `plugins.enabled`** at config schema v21+. So even when the `enable` call fails, dropping the files can still result in enablement later.
- `installHermesHooksIfNeeded()` doesn't check whether Hermes exists. Every `agentdeck daemon install` creates `~/.hermes/plugins/agentdeck-observer/` on machines that have never had Hermes, and a user who installs Hermes afterwards inherits an observer they never agreed to.

I want to be fair about the precedent here: our OpenCode installer also writes unconditionally, so that part is house style. The difference is that OpenCode has no consent allow-list — dropping the file *is* the mechanism. Hermes has one, and it is documented as a security boundary.

## 2. The better path: `hooks.outbound`, not a plugin

Hermes ships an integration built for exactly this, and I think it dissolves most of the PR's problems rather than patching them. From the hooks documentation:

```yaml
hooks:
  outbound:
    - name: agentdeck
      url: http://127.0.0.1:9120/hooks/hermes_session_end
      events: [on_session_end]
      timeout: 10
```

Its stated use cases are literally ours — *"Notify a CI system or dashboard when an agent turn finishes (`on_session_end`)"*, *"Feed tool activity into external monitoring (`post_tool_call` with a `matcher`)"*. Point by point against the current design:

- **"No consent prompt. Outbound targets execute no code on your machine"** — §1 disappears entirely. No files in the plugin directory, no allow-list to bypass, no `execFileSync`, no grandfathering exposure.
- **"Fire-and-forget, off the hot path. Events are serialized and queued instantly; a single background thread performs the HTTP POSTs. A slow or dead endpoint can never stall a tool call or an agent turn."** — this removes the blocking `urlopen` your `pre_tool_call` currently performs on the critical path. Worth noting our own convention for the Claude hooks is a split bound (`--connect-timeout 0.2 --max-time 0.8`); `urlopen(timeout=0.8)` is a single bound, and if Hermes dispatches plugin hooks on an event loop, a synchronous urlopen stalls the loop rather than one call.
- **`delivery_id` + `timestamp` in the signed body**, with documented dedupe-and-reject-stale semantics, gives a principled answer to the duplicate-event problem in §4.
- **The wire format includes `cwd`** (`{"hook_event_name": ..., "session_id": ..., "cwd": "/home/user/project", "extra": {...}}`). The plugin-hook payload catalog has no `cwd` field on any event — see §5 — so webhooks are the only path on which your `projectFrom()` can actually produce a project name.

There is a pleasant coincidence: `HermesSessionObserver` already reads `session_id`, `cwd`, and a nested `extra` object via `nestedField()`. That is the outbound wire shape almost exactly. The bridge-side observer may need very little change — it's the delivery mechanism that's wrong, not the receiver.

The costs are real but small: it's a merge into the user's `config.yaml` (same class of operation as the `~/.codex/config.toml` merge we already do, and it needs the same care about preserving user entries), and *"Changes take effect on the next CLI session / gateway restart."*

## 3. Blocking regardless of mechanism: `observed:hermes:` is missing from four hand-mirrors

New `observed:<agent>:<uuid>` ids have to be registered in every copy of the prefix list, and there are four:

```
shared/src/session-utils.ts:25                       claude|codex|codex-app|opencode|antigravity
apple/.../Daemon/Server/DaemonServer.swift:5831      claude|codex|codex-app|opencode|antigravity
apple/.../Daemon/Timeline/DaemonTimelineStore.swift:243   claude|codex|opencode|antigravity
android/.../state/TimelineDisplay.kt:392-396         (five entries)
```

None are updated, so `rawSessionId('observed:hermes:<id>')` returns the string unchanged, and `sameSession()`, focus resolution, and `resolveSessionIdPrefix()` for the 31-character ids devices report all break for Hermes rows. CLAUDE.md documents this specific trap ("four copies drifted into three different agent lists"), and note that the third line above is *already* missing `codex-app` — the drift is live, not hypothetical. If Hermes lands, adding a fifth agent to an ungated four-way hand-mirror is the wrong shape; that list should become generated first.

## 4. Two `hermes_stop` events per turn

You register both `post_llm_call` and `on_session_end` to `hermes_stop`. Upstream's shipped hook catalog:

- `post_llm_call` — *"Successful, non-interrupted turn finalization"*
- `on_session_end` — *"Canonically at each turn finalization"*

Both are per-turn, so every turn emits two stops. The observer is idempotent here so nothing visibly breaks, but it's double traffic on a synchronous path.

Related: `pre_llm_call` has two shapes upstream, and the official langfuse plugin guards against exactly this with `if not isinstance(messages, list): return` to avoid an extra trace per turn. Without that guard, `hermes_user_prompt_submit` fires more than once per turn too.

## 5. Smaller things

- **No `cwd` on any plugin hook.** The payload catalog gives `on_session_start` only `session_id`, `model`, `platform`; `pre_llm_call` adds `user_message`, `conversation_history`, `turn_id`, `parent_session_id`, `sender_id`. So on the plugin path `projectFrom()` always falls through to `Hermes · <platform>` and never names a project. (On the webhook path it works — §2.)
- **The brand-mark guard covers one surface of four.** `agentLogoIcon('hermes') === ''` protects `shared/src/svg-renderers/agent-logos.ts`, but Swift, Kotlin and the ESP32 firmware each carry their own agent→glyph maps that fall back to the OpenClaw creature. So the substitution the PR set out to avoid still happens on three surfaces, and the test only pins the one that doesn't.
- **`HERMES_BIN` fallback is `~/.local/bin/hermes` with no PATH lookup**, so Homebrew and pipx installs never match and always take the warning branch.
- `writeFileSync` here vs `writeTextAtomic` in the OpenCode installer — worth matching.
- `pre_llm_call`'s `user_message` becomes `currentTask`/`goal` and travels to every paired device over plaintext LAN WebSocket. `goal` already does this for Claude and Codex so it isn't a new class of exposure, but it should be deliberate.

## Where this stands

The engineering is sound and the upstream reading is accurate — my objections are about the delivery mechanism and about what a fifth observed agent costs us in mirrors, not about the quality of the work.

But I owe you honesty about the product question, which is genuinely open on my side. AgentDeck currently has two integration tiers: managed/observed coding agents in a repo (Claude Code, Codex, OpenCode), and OpenClaw, which gets a full Gateway RPC adapter and is steerable. Hermes sits in the same category as OpenClaw — an always-on personal agent gateway with messaging platform adapters — and I haven't decided whether a second one belongs, whether it should be observe-only or eventually steerable, or whether the two would ever run side by side on one machine. That's not a question this PR can answer for me, and it would be unfair to let you rebuild it twice while I work it out.

So: I'm not asking for a revision right now. If you'd like to keep momentum on something that lands regardless, §3 stands on its own — turning the four-copy observed-prefix list into a generated SSOT with a drift gate is a real cleanup that we need whether or not Hermes ever ships, and it would already be paying for itself today given the `codex-app` drift. I'd take that PR.

I'll follow up here once I've decided on the Hermes direction.

puritysb:
Closing this implementation branch as superseded, rather than rejecting the Hermes idea. The plugin-based delivery path conflicts with Hermes's consent boundary, the product tier is still undecided, and this branch now conflicts with current `master`. If Hermes integration proceeds, it should start as a fresh PR around the consent-safe `hooks.outbound` path and the generated observed-prefix SSOT discussed in review. There is no need to rebase or continue this branch.

PR #176, opened 2026-08-12
fix(bridge): preserve multi-byte UTF-8 across stdin chunk boundaries

uto-usui:
## Problem

Pasting text that contains multi-byte characters (Japanese/CJK, emoji) into a session started with `agentdeck claude` corrupts roughly one character per stdin chunk boundary. The damage is visible in the agent's input field before anything is submitted, and it is permanent — the original bytes are gone.

Example, pasting a Japanese line into Claude Code through AgentDeck:

```
expected:  ...に残した検証用 Bun バイナリの削除です。
actual:    ...に残した検<?><?>用 Bun バイナリの削除です。
                        ^^^^^^ 証 (E8 A8 BC) replaced by U+FFFD
```

## Cause

`PtyManager.attachTerminal` decodes each raw stdin chunk on its own:

```ts
stdin.on('data', (data: Buffer) => {
  if (this.ptyProcess) {
    this.write(data.toString());   // stateless decode, per chunk
  }
});
```

`stdin` is in raw mode and has no encoding set, so `data` is a `Buffer` whose boundaries fall on arbitrary byte offsets. A UTF-8 character that straddles two chunks is therefore decoded as two separate invalid sequences, and each leftover byte becomes `U+FFFD`.

The damage has a distinctive signature. A 3-byte character split as `E8 | A8 BC` yields **three** replacement characters (one per byte, all invalid on their own); split as `E8 A8 | BC` it yields **two** (a truncated sequence plus a stray continuation byte). Both appear in practice, which is what identifies a boundary split rather than bad input.

Measured over ~17,800 real prompts from one user's history, corruption tracked paste size exactly as a per-boundary failure predicts:

| paste size | corrupted |
|---|---|
| < 256 B | 1.8 % |
| 256 B – 1 KB | 42–68 % |
| 1 KB – 2 KB | 90 % |

ASCII input is unaffected, since every character is one byte and no split can damage it.

## Fix

Set a stateful decoder on stdin. Node's `setEncoding('utf8')` keeps the trailing bytes of an incomplete sequence and prepends them to the next chunk, so characters survive any boundary:

```ts
stdin.setEncoding('utf8');
```

`data` then arrives as a `string` and is forwarded as-is. No behavior changes for single-byte input, and no protocol layer is involved — bracketed-paste markers and escape sequences pass through unchanged, as they did before.

## Verification

Reproduced and confirmed against a real session, driving `agentdeck claude -c <recorder>` over a PTY and sending a correct UTF-8 paste from the parent. The recorder uses `setEncoding('utf8')` itself, so any `U+FFFD` it observes was introduced by the bridge:

```
before:  U+FFFD count = 7   (run lengths [2, 3, 2])
after:   U+FFFD count = 0   (byte-for-byte identical to what was sent)
```

Added `bridge/src/__tests__/pty-manager-stdin.test.ts`, which feeds a Japanese string through `attachTerminal` split at offsets that fall inside characters, and asserts the forwarded text matches the original. It fails on `main` and passes with this change.

## Notes

This is a one-line behavioral fix in the stdin path; the `data: Buffer` → `data: string` type change and the removal of `.toString()` follow from it.

Checks run locally: `vitest run bridge/src/__tests__/{adapter,pty-manager-stdin}.test.ts` (95 passed) and `pnpm --filter @agentdeck/bridge typecheck` (clean).

## AI assistance disclosure

Per CONTRIBUTING, disclosing this: the investigation that located the bug and the patch itself were produced with an AI coding assistant (Claude Code). I reviewed the change, and the before/after verification above was run against a real `agentdeck claude` session on macOS.


puritysb (approved):
Verified the stateful UTF-8 decoding fix preserves split multi-byte input while leaving ASCII/control input unchanged. Local targeted tests, full 2,908-test suite, build, typecheck, and the latest GitHub CI/Design System checks all pass.

PR #177, opened 2026-08-12
Add Apple usage analytics and review prompts

puritysb:
## Summary

- request the native StoreKit rating prompt only after meaningful live-session use on three distinct days, with production-receipt gating and a 180-day local cooldown
- add a user-initiated App Store review link in Settings
- add `pnpm analytics:apple` to download Apple's opt-in, privacy-thresholded aggregate App Usage reports without adding an analytics SDK or AgentDeck-operated backend
- document the privacy boundary, App Store feature matrix, review notes, release operation, and current GitHub triage state

## User and developer impact

The app can ask engaged App Store users for a rating at a restrained cadence while keeping all eligibility state local. Maintainers can inspect Apple's aggregate App Store Connect engagement reports using existing API credentials; no session identifiers, project names, commands, device identifiers, or custom analytics events are uploaded by AgentDeck.

## Validation

- `pnpm test` — 179 files, 2,910 tests passed
- `pnpm build`
- `pnpm docs:check`
- `pnpm design-system:check`
- `node --check scripts/app-store-connect-analytics.mjs`
- `git diff --check`
- macOS Debug build and iOS Simulator build recorded in `DEVELOPMENT_LOG.md`
- AppReviewPromptPolicy standalone runtime assertions and App Store submission validator recorded in `DEVELOPMENT_LOG.md`

## Known local test-host limitations

The new Swift test file compiles for both test targets. Full local macOS XCTest launch is blocked by an Xcode 26.6 LaunchServices assertion when the installed App Store app has the same bundle ID, and the iOS test target has a pre-existing `SpokenDigestParityTests` reference to the iOS-excluded `DaemonServer`. GitHub Apple checks remain the authoritative clean-run validation for this PR.


puritysb:
Self-review found three items to address before marking this PR ready:

1. **[P1] `init` and the documented API-key role disagree.** `init` sends `POST /v1/analyticsReportRequests`, but Apple permits only an **Admin** key to create/manage report requests. `RELEASING.md` currently says the key needs Sales and Reports, which can download existing reports but will receive 403 for `init`. Document separate Admin setup vs download roles (or use an Admin key only for `init`).

2. **[P1] Existing usage history is not requested.** The only creation path makes an `ONGOING` request. Apple documents that historical backfill requires `ONE_TIME_SNAPSHOT`; `--days 30` filters processing dates and cannot recover the 30 days before the request existed. Since the goal is to understand the users already acquired, add a snapshot command/path and make the processing-date semantics explicit.

3. **[P2] The rating prompt can interrupt active work.** `ContentView.considerAppStoreReview()` requires only an alive session, and its task id contains session IDs but not session state. A session can remain `processing` or `awaiting_*` for the two-second delay and still trigger the prompt. Gate the final request on an idle/completed transition or another natural stopping point, consistent with Apple's ratings guidance.

References: [Analytics Reports roles and access types](https://developer.apple.com/documentation/AppStoreConnectAPI/downloading-analytics-reports), [Apple ratings and reviews guidance](https://developer.apple.com/design/human-interface-guidelines/ratings-and-reviews).

All five CI checks pass and GitHub reports the PR as mergeable; these are product/behavior findings rather than CI failures.

PR #201, opened 2026-08-16
Enterprise posture devlog + let ADB reverse survive --loopback

puritysb:
Two follow-ups to the enterprise network posture work merged in #198:

**1. Development log entry.** #198 merged three workstreams but only two got a log entry; the posture work (`--local`/`--loopback` axes, "off is derived", Pixoo auto-discovery default flip and its two regressions) is the one a future session is most likely to grep for.

**2. ADB reverse now survives `--loopback`.** The posture originally turned the ADB reverse tunnel off with the LAN emitters, on "a security posture defaults to off". But `adb reverse` passes exactly the test that kept USB serial alive: it rides the USB cable and terminates on the host's **own loopback** — the very interface this posture binds — and puts nothing on the LAN. Turning it off silently killed every USB-tethered Android dashboard (three in the reference fleet, no token pairing, tunnel-dependent) for zero security gain. `--local` (no device modules at all) still turns ADB off. The startup posture line now names the USB channels that stay on.

Docs updated in the same pass: `docs/daemon.md` (table + "what you give up"), `docs/cli.md`, `CLAUDE.md`, roadmap item 6 status, and the devlog entry itself records the reversal.

Tests: `network-posture.test.ts` updated (18 pass) — loopback keeps `adb: 'auto'`, the derived-off shape test now allows exactly the two USB channels, and the posture-line test pins the "stay on" wording.


puritysb:
Superseded: both commits here were absorbed into master by #202's squash merge (its branch was cut from this branch's head), so this PR's diff against master is empty. The adb-over-WiFi transport gap found in review will be fixed in a follow-up PR.

PR #262, opened 2026-08-23
Make an OpenClaw approval answerable

puritysb:
Opened on behalf of the session that wrote it (`agentdeck-2b`), which ended before pushing. Two commits, rebased onto `7c57e501`.

## The measurement

An exec approval reached every deck as a bare shell command with no statement of what was being asked or why. Measured 2026-08-23: **8 real approvals, 7 closed unanswered after 75–402s.** The data was never missing — it was dropped at the surfaces.

## Three defects, one shape each

- **The deck cut the question to 18 chars from the front.** These are shell commands, so a head cut keeps the verb every request shares and drops the object that identifies this one: `sed -n '20,35p' ~…`. `summarizeQuestionForKey` keeps verb + object basename instead (`sed … config.py`), scanning forward so the object belongs to the verb it kept. `rm … build-cache` is why this is a safety rule, not a legibility one.
- **The reason approval was demanded had no field to ride in.** `parseExecApprovalRequest` already extracted it; `injectOpenClawSession` dropped it. Adds `SessionInfo.questionDetail` (additive) carrying the policy warning, the cwd, and *which* OpenClaw session asked — `main`, a cron job and a model-eval run all arrive as the literal "OpenClaw", which is how an approval raised by an eval harness looked like it came from a conversation the user could not find.
- **The D200H rendered zero pixels of subject.** `question` was resolved only as a press echo and `renderDetailInfo` had no question parameter, so three live decision keys sat over no statement of what they applied to.

It also splits `abandoned` from `denied`: every non-approval closure wrote `denied`, so a deliberate refusal was byte-identical to a prompt nobody ever answered — and the second is the common case, and the one that means the prompt failed to reach a human. That surfaced a latent bug where `tool_resolved` returned `success` regardless of status, drawing a green check on denials.

## Second half — the timeline was a biased sample, not a log

OpenClaw is Gateway-observed, and that stream carries the prompt, the response and the approvals it must block on — **never the tool calls**. The only tools that ever reached the timeline were the ones that happened to need approval. OpenClaw writes a full per-session transcript anyway, so `openclaw-transcript-timeline.ts` + `openclaw-timeline-feed.ts` read it and emit the complement: `tool_exec` rows only, never a duplicate of the live stream.

Bounded three ways: only sessions inside a 30-minute activity window (the store keeps every session forever — 103 on the machine this was built against); a first sighting emits **nothing** (deliberately the opposite of `KiroTimelineFeed`, because OpenClaw has no blank-strip problem and seeding would inject stale eval-run tool calls); and rows attribute to the existing virtual `openclaw-gateway` row rather than minting sessions.

Every transcript fixture is copied verbatim from a real store — including a failure reported by `isError` alone with no exit code, which a reader trusting the exit code scores as success.

## Third half — the feed was on, and had never published a row

The item above ("live emission has not been observed yet") turned out to be
load-bearing. "OpenClaw was idle" was not why nothing was seen; it was why
nobody looked. Measured end to end against a dedicated probe session
(`agent:main:agentdeck-probe`): **three consecutive OpenClaw turns each wrote a
tool call to the store, the reader returned it correctly every time, and the
timeline held zero `tool_exec` rows.** Two independent defects, either of which
alone is total.

**1. The producer was driven by a signal its own subject cannot move.** The
pump rode `passiveSessionObserver.onRefreshed`, which fires only when the
observed *process* set changes. Every other transcript feed may ride that edge
because its subject is in that set — a Kiro turn moves its own row's
`state`/`lastActivityAt`. OpenClaw is reached over the Gateway and has no
process row at all, so an OpenClaw turn moves nothing the observer can see. The
daemon's own log: the pump ran **exactly twice, both at startup** (empty cache →
populated always counts as a change) and never again. It now has its own timer.

**2. The only type this feed emits is the type eviction sheds first.** With (1)
fixed the feed logged `emitting 1 row(s)` and the strip was still empty — the
entire 200-row buffer held no `tool_exec` from *any* agent. `addEntry` pushes
then evicts, and `evictOne` takes the oldest plain `tool_exec` first, so on a
full buffer (the steady state on any busy machine) the row just added was the
only one present and **evicted itself on insertion**. OpenClaw is carved out of
that rule the way subagent dispatch rows already are — in **both** daemons,
because the feed is Node-only but `timeline.json` is owned by the two in turn,
and a one-sided carve-out leaves the replay buffers diverged.

Both failure modes look identical from outside — a tick that emits nothing
writes nothing — which is why the emit path is now logged at all.

Measured after both fixes, on the shipped (non-debug) daemon: the probe turn
logged the emission and the row landed, `session: agentdeck-probe` leading its
detail. Rows now accumulate and survive in the live buffer.

## The two remaining items, now decided

**Trajectory → APME: no.** The premise was wrong. Across **all 484 trajectory
files** in this store only 8 event types ever occur and **none is a tool
event** — `<uuid>.trajectory.jsonl` carries turn boundaries
(`session.started` / `prompt.submitted` / `model.completed` / `session.ended`,
one `runId` per turn) plus per-turn provider/model and token usage, but no
steps. A `SessionSample` built from it would have an empty trajectory; the tool
calls live in `<uuid>.jsonl`, which is the file this feed already reads. So the
Gateway stream stays the owner of turn/task boundaries — it is the only push
channel and already owns them, while trajectory is polled and would arrive after
the fact, adding latency and the double-count without adding depth. What
trajectory uniquely has is per-turn *identity and cost*, which is an enrichment
axis, not a boundary axis: if wired later it should enrich existing APME turns,
never open and close runs.

Surfaced while deciding this, and left untouched because it changes what a
"run" means in existing eval data: **the OpenClaw APME run is scoped to the
Gateway connection, not to a session.** `apmeSessionId` is a single value set
once at connect, so the user's chat, the cron heartbeat and every concurrent
eval run land in one run with one model field, with idle-gap heuristics for turn
boundaries. For an eval workload that is the defect worth fixing, and it should
be settled before any trajectory work.

**Swift daemon mirror: deferred.** Reaching `~/.openclaw` from the sandbox needs
the security-scoped bookmark consent flow Kiro uses — a Settings flow, not a
code port. And this round is the argument for not porting first: the Node path
hid two total defects until it was measured live, and mirroring unvalidated
behaviour onto a surface that is harder to instrument (App Store build, no debug
log) would just move the blind spot. **The eviction carve-out did land on both
sides**, since that rule cannot diverge.

## Verification

vitest **222 files / 3455 tests**, Swift `TimelineTests` **74 tests**, macOS
build succeeded. Both new regression gates were checked by *removing* the
carve-out and confirming they go red — the Node one against the pre-fix `dist`
as a control, the Swift one by reverting the predicate (`TEST FAILED`) — rather
than trusting a green suite.


puritysb:
**Verification — re-run independently before publishing, not carried over from the authoring session's report.**

Ran in a detached worktree at `3d3fdc09`, not in the shared tree:

| Gate | Result |
|---|---|
| `pnpm build` | clean |
| `pnpm -r typecheck` | 4/4 packages Done |
| `pnpm vitest run` | **222 files, 3452 passed, 1 skipped, 0 failed** |
| `check-preview-mirror-sync` | 10 pins / 6 mirrors in sync |
| `verify-tokens-sync.py` | 99 tokens × 7 mirrors in sync |
| `check-docs` · `check-surface-mirrors` · `sync-pages-nav --check` · `design-system --check` · `generate-esp32-board-matrix --check` | all pass |
| `xcodebuild test -scheme AgentDeck_macOS` | **TEST SUCCEEDED** (local, full suite) |

The macOS build was run locally on purpose: `tsc` passing says nothing about whether the Swift mirrors compile, and this change edits six Swift files including the generated protocol and the approval-rules generator's output.

CI on this head agrees: **CI · Apple Tests · Android Tests · Design System · ESP32 Sim Compile — all success.**

Two things worth noting about *how* the mirror pins got here, since this is the first branch to land after the gate started running at all. `Preview mirror sync check` sits after the vitest step in the `test` job, and master's vitest step had been failing, so the gate had never executed on master until #258/#260 cleared the way. Both pins this branch moves (`session-slot-renderer.ts`, `d200h-layout.ts`) drifted on **pure additions** — file-level hashing means "I only added lines" is not an exemption — and both were recomputed with `git hash-object` from the resolved files rather than copied out of a CI message.

Not verified here, and stated in the PR body rather than implied: the feed's **live** emission has not been observed running. It is unit-tested against fixtures copied verbatim from a real store.

PR #263, opened 2026-08-23
Subscribe to the OpenClaw events APME was waiting for

puritysb:
## What was measured

| | |
|---|---|
| OpenClaw runs recorded | 665 |
| **Runs holding zero turns** | **657 (98.8%)** |
| Last turn APME ever recorded | 2026-08-16 |
| Every turn that did exist | came from a prompt typed *through AgentDeck* |

Nearly every run is exactly ~1800s long — that is the 30-minute orphan reaper closing an empty shell.

## Why

The Gateway's `chat` frame is **assistant-only**. Verified against the emitter itself (`openclaw@2026.7.1-2`, `dist/server-chat-wgxNCdC3.js` → `emitChatDelta` / `emitChatTerminal`): it carries `message:{role:'assistant',…}` and no prompt, no `tools`, no `modelId`, no tokens. So nothing ever emitted a `turn_start`, and a `turn_response` with no open turn is dropped by `Collector.setTurnResponse` (`if (!turnId) return`).

**The fix is one RPC we never sent.** `sessions.subscribe` takes no params and runs under `operator.read` — the scope `sessions.list` already uses — and it is the entire admission gate for `session.message` and `session.tool`: the Gateway's recipient set for both is `sessionEventSubscribers ∪ sessionMessageSubscribers`, and a connection in neither set receives nothing at all. The handlers for both events (`openclaw.ts`) and both parsers (`openclaw-hook.ts`) had been wired since they were written **without ever receiving a frame**. The Swift adapter already subscribed — Node was the outlier.

## Three copies of one guess

`ChatEventPayload` declared `prompt`/`tools`/`modelId`, `chat-final-with-tools.json` supplied them, and both parity suites read them back. A fixture, a type and an assertion authored together from one guess agree with each other forever; nothing in the loop had been compared against OpenClaw. `session.message` was typed as a flat `{role,text}` and `session.tool` as a flat `{name,status,input,output}` — shapes the Gateway has never sent. `exec-approval-requested.json` still carried the flat shape that `shared/src/openclaw-approval.ts` documents as disproven, including the `'allow'` decision the Gateway rejects.

Fixtures are now **captured off a live Gateway** and content-scrubbed, and the parity tests assert on the **parser's output** rather than restating fixture fields.

## Run scoping had to land in the same round

Once data flows, a connection-scoped run interleaves the user's chat, every cron job and every eval run (a suite opens a fresh key per run) into one trajectory under one `model_id`, with the idle-gap boundary computed across unrelated conversations. Runs are now `openclaw:<sessionKey>`, opened lazily per key, **each with its own idle-gap timer** — a heartbeat firing every few minutes must not keep resetting the boundary of a chat that ended half an hour ago, and a task that never closes is never evaluated.

Old `openclaw-<uuid>` rows are **not backfilled**: the session key was never recorded on them, so any attribution would be invented.

## Two defects only visible once capture worked

- **OpenClaw was the only agent leaving turns open under a closed task** (measured across all agents). `chat.final` is its stop signal, so it now emits `turn_end` (`end_source='stop'`; a cancel gives `interrupted`). The `turn_end` span kind was a no-op with no emitters, so filling it in affects no other adapter.
- **`tasks.first_turn_index` was persisted NULL for every agent** — 294 of 1080 closed claude-code tasks. The row is INSERTed from inside the turn_start handler (before the turn exists) and `closeTask` only ever wrote `lastTurnIndex`. `reapAbandonedRun`'s `COALESCE` was this defect's workaround, not its fix.

## Swift mirror

It subscribed but read `session.tool` flat, so `isPlaceholderOnlySessionTool` discarded **100% of real tool frames as noise** — a drop whose name made the loss look deliberate. Its tool status also came from the frame's top-level `status`, which belongs to the session snapshot, not the tool.

## `agentdeck daemon restart`

It reported `Daemon restarted (PID …)` from `spawn()` returning a pid, which is not evidence a daemon started — a child dying on EADDRINUSE was announced as success while the reason sat unread in the daemon log. It now waits on `waitForDaemonExit` + `waitForPortBindable` (the fixed 1500ms sleep was a guess; macOS holds a NECP reservation ~14s after a listener is cancelled), then verifies `/health` reports the pid it spawned. The pid comparison is load-bearing: probing the port alone is satisfied by the daemon it was trying to replace.

## Verification

End to end against a live Gateway, two session keys:

```
openclaw:agent:main:agentdeck-probe    glm-5.2  turns=1  steps=3
openclaw:agent:main:agentdeck-probe2   glm-5.2  turns=1  steps=1

trajectory: user_message → model(54/23 tok) → tool(read, success, args+result)
            → model → assistant_message
turns:      end_source=stop, duration 9s, tool_calls=1 on the tool-using turn
tasks:      each key closing on its OWN idle gap; first/last_turn_index = 0..1
restart:    "Daemon restarted (PID 18227) on port 9120" → /health pid 18227
```

Both new gate sets were checked red: run against the previous parsers, all four real frames yield `[]`; reverting the Swift `data` lookup fails 8 assertions.

`vitest 3479/3479` · `Swift 674/674` · `docs:check` · `verify-version` clean. Design lint untouched (no file in this change is flagged).

## Follow-ups, deliberately not in scope

- `chat_start` rows still read `자동 작업` for gateway-originated turns. The real prompt is now available from `session.message`, but writing it into `lastPrompt` also moves the `chatIsAutomated` heuristic and the 8h automated-dedup.
- The timeline's tool rows stay owned by the transcript feed; `session.tool` is routed to APME only, so the two do not double-emit.
- **trajectory → APME is closed as not-needed** (the deferred item from the previous round): the per-turn provider/model/token usage it was wanted for now arrives first-hand on `session.message`(assistant) as a `session_meta` span. Reading the same columns from a polled file too would be a second source for a value we already own.


puritysb:
### Two follow-up commits after the initial push

**`3fcb904f` — the client timeline mirror had drifted.** `TimelineStore.evictOne` names itself a port of `DaemonTimelineStore.evictOne`, but never got the OpenClaw carve-out: the daemon kept those `tool_exec` rows under buffer pressure while the app's own view shed them. Only reachable now, since the Swift adapter emitted no OpenClaw tool rows at all until it was taught the real `data`-nested shape.

Worth flagging how the test went: my first version filled the buffer with `tool_exec` rows and **passed with or without the carve-out**. The bug bites only when the OpenClaw row is the *only* plain `tool_exec` present, so it is the first match and evicts itself on insertion. With chat rows as filler it fails without the carve-out.

**`6f2b427f` — live traffic caught a defect this round introduced.** After deploying the per-key split, a real `agent:main:main:heartbeat` key opened a run holding **0 turns, 0 steps, 0 events**. The run was opened while *building the context*, and a context is built for every frame — including ones that parse to zero spans. The Gateway filters heartbeat message pairs out of `session.message`, so frames that record nothing are routine. That is the empty-run noise this PR is about, reintroduced one layer down.

A key's run now opens on its **first ingestable span**. The idle-gap timer resolves at fire time and never opens a run — a boundary closing a task that was never opened would mint an empty run at the exact moment a conversation went quiet.

### One thing the PR description overstates, now corrected in-tree

The connection-scoped **fallback** run is still opened eagerly at connect. Since essentially every frame carries a session key, that row is usually empty and reaped as an orphan — **one row per Gateway connection, the same rate as before the split**. Making it lazy means restructuring `setApmeSession`, which `hasDirectApmeIngestion` gates the bridge's timeline→span conversion on, so I left it and wrote it down in `CLAUDE.md` + the devlog rather than implying it away.

Also corrected: "session.tool is routed to APME only" is true for **Node** (its timeline tool rows come from the transcript feed), but Swift has no transcript feed and `session.tool` **is** its timeline source.

`vitest 3480/3480` · `Swift 675/675`. Both new gates checked red against the pre-fix code.

puritysb:
**Adversarial review — two independent reviewers plus a direct pass. Two merge-blocking findings, both fixed in `4de4b88d`.**

### M1 — the Swift daemon dropped every assistant `session.message` (CONFIRMED)

`emitTimelineEntry(fromSessionMessage:)` read `payload["role"]` / `["text"]` / `["content"]`, but the frame's top level is the session snapshot — the message is nested. Verified against **this PR's own captured fixtures**: top-level `role`/`text`/`content` absent in all three; `message.content` is a `String` for a user message and an **array of typed blocks** otherwise. The lone `as? String` yielded `""` for every assistant frame and the empty-text guard discarded it — reading as noise suppression while throwing away 100% of real replies. The same shape as the `session.tool` body bug this PR fixes one function up.

`role` never resolved either, so `model_response` was unreachable and every row was a `model_call`. Per this PR's own CLAUDE.md, `session.message` is the response's owner for runs the Gateway hides from the control UI — on the App Store daemon that fallback did not exist, so such a run recorded a prompt with no answer.

### M2 — the Swift parity suite could not fail on adapter behavior (CONFIRMED)

`GatewayParityTests` loads the fixtures and asserts on their own dictionaries. Reverting the adapter fix leaves all 8 green. It documents the shape without exercising anything that reads it — and the two fixtures the adapter mishandled had **no Swift assertion at all**. This is the anti-pattern the fixture README itself condemns ("a fixture earns its place by driving the real parser"). New cases in `OpenClawToolNoiseTests` drive the real parser.

### D1 — `phase: 'error'` got opposite verdicts on the two daemons (CONFIRMED, fixed)

Node derived status from `isError` alone, so a frame whose *phase is* the failure signal recorded as a **successful** tool result — the wrong side of the disagreement, and it feeds the tool tally the APME judge scores. Both daemons now classify the same three terminal phases. Swift also no longer returns an unrecognised phase verbatim into `TimelineEntry.status`.

### Mutation results, including the one that did not have teeth

Reverting the fix turns two new Swift tests red. `testErrorPhaseIsTerminalAndFails` passes **either way** — pre-fix Swift returned the phase verbatim and so answered `"error"` too. Rather than delete or quietly keep it, it is labelled in-file as a contract pin, and the gate that actually fails without the fix is the Node one added in `apme-sample-timeline.test.ts`.

### Independently verified against the vendor, not taken on trust

The PR's protocol claims were checked against the installed `openclaw@2026.7.1-2` source: `sessions.subscribe` exists at `scope: "operator.read"`; the `session.message` recipient set is literally `sessionEventSubscribers ∪ sessionMessageSubscribers`; the payload nests under `message`; `emitChatDelta` carries assistant text only. **The fixtures are captured, not composed.**

Live APME database confirms both the premise and the effect: 673 OpenClaw runs with only 12 carrying turns, and post-deploy session-key runs now carry real turns (1–2 each). The disclosed eager connection-run gap is real and measurable — one empty run one second after daemon start — and is documented rather than implied away.

### Verification

Swift 680 tests green · vitest 3481 green · preview-mirror 10/10 · docs, design-system, protocol drift green · CI green on `4de4b88d`.

### Left as follow-ups, not fixed here

- Swift still has dead `chat`-frame `prompt` readers whose comment now states a falsehood; `automatedRunId` is assigned only inside that dead branch, so a cron run's response cannot be flagged `automated` on Swift.
- `agentdeck.turn_end_source` is typed `string` rather than `TurnEndSource` at the one site that writes `turns.end_source`.
- `isHeartbeat` is declared in the SSOT and read by nobody.
- Swift's tool `input`/`output`/`name` resolvers still prefer top-level over `data`; the fixtures cannot rule out a snapshot key collision because the snapshot was trimmed at capture.

PR #264, opened 2026-08-23
Accumulate per-message usage, and close the turn the tests never checked

puritysb:
**Follow-up to #263.** I merged that PR on one reviewer's findings; the second reviewer's report arrived afterwards with five more, and these are the ones that record wrong data or invent a signal. My error on the sequencing — the findings below were real and reviewable before the merge.

## B1 — per-message token counts fed to a cumulative API (data corruption)

`updateUsage` assigns `input_tokens` to the run row **absolutely** and derives its per-task ModelEvent as `max(0, cur - prev)`. Its only previous caller is the PTY poller's running `UsageTracker` snapshot. A `session.message`(assistant) frame reports **one message's** tokens, and those are not monotonic — measured on a real OpenClaw session:

```
in=30302  in=21375  in=335  in=96  in=114
```

| message | run row after | ModelEvent delta |
|---|---|---|
| 30302 | 30302 | 30302 |
| 21375 | **21375** ← total goes *down* | 0 |
| 335 | 335 | 0 |
| 96 | 96 | 0 |
| 114 | **114** | 18 |

Final run row: `input_tokens = 114` for a session that consumed **52,222**, and `recomputeSampleCost` prices from what survives. The dedup key `${curIn}:${curOut}` also assumed cumulative values are unique.

`addUsageIncrement` keeps the running total and hands `updateUsage` the cumulative series its arithmetic is written for.

## The `turn_end` close had no gate

Reverting the span **emission** goes red. Reverting the collector's `closeTurn` left **all 2417 tests green**. The reverted half is the load-bearing one: it writes `turns.end_source` — the column `agentdeck apme stop-health` reads, and the column CLAUDE.md says exists *because* a dropped close is otherwise byte-identical to a normal one. Now driven end to end.

## B5 — a trailing newline manufactured a Stop loss

Two producers open a turn for one prompt: our `chat.send` span carries it verbatim; the Gateway's `session.message`(user) echo arrives trimmed, envelope-stripped and capped at 8,000 chars. The duplicate-open guard required **exact** equality, so `"fix this\n"` vs `"fix this"` missed — and a miss closes the open turn via `resolveDisplacedTurnSource`, which for a non-`claude-code` agent returns `next_prompt`: **the unrecovered-Stop-loss bucket**. A stray newline invented a Stop loss for an agent that has no Stop hook, stranded an empty turn, and shifted every later `turn_index`.

Now compared trimmed, with the truncation tolerance `askEchoMatches` already uses — a prefix counts only when the shorter side is *at* the cap, so two prompts sharing an opening line still read as different.

## C1/C2 — `daemon restart` gave up while the child was still waiting

The parent budgeted 8s for the port and 20s for verification, but on EADDRINUSE the child waits `PREFERRED_PORT_RECLAIM_MS` (20s) before falling back — and the comment in that same hunk names the ~14s NECP reservation it was budgeted under. A contended restart printed `restart FAILED` and exited 1 *as the daemon came up*; the natural retry then kills it. Both budgets now derive from the child's worst case. The failure message also names stdout, since the likeliest cause (`daemon start`'s incumbent guard) reports through `log()`.

## B2 — a decision resting on a capability that does not exist

CLAUDE.md justified not reading the transcript partly because provider "is already ingested". It is not — `gen_ai.system` is emitted and declared with **no consumer** (`grep` finds the emitter and the type, nothing else). The claim is corrected rather than the field quietly dropped, since its destination (`shared/src/model-provider.ts`) already exists.

## Verification

Every new test mutation-verified — reverting each fix turns its own gate red: `expected 114 to be 52222`, `expected null to be 'stop'`, `expected 2 to be 1`. vitest **3484** green, preview-mirror 10/10, docs green.

## Knowingly not fixed here

- **B4** — `session.message` is a projection capped at 8,000 chars with envelopes stripped, so for a long reply whichever of `chat.final` / `session.message` lands second wins and can overwrite `turns.response` with the truncated text. Which lands second is inferred from code, not measured; needs a live capture before choosing an owner.
- **B3** — per-turn model still lands run-level via `updateModel`, which is what the emitter's own comment calls the bug. Fixing it means a `turns`-level column.
- Provider ingestion (B2's other half), `messageSeq` gap detection, `openRun` caching a failed insert as success, and the dead `toolCallBlocks` / `apmeSessionIds`.


puritysb:
**Four review rounds. Merging on a clean fourth.**

| round | target | findings |
|---|---|---|
| 1 | original #263 | 5 merge-blocking |
| 2 | **the fix** `3b5963b0` | 3 — including **a new defect the fix introduced** |
| 3 | **the fix's fix** `02e5497a` | 0 blocking; 1 downstream consequence |
| 4 | `3805c2c9` | 0 blocking; 1 convention violation |

Rounds 1 and 2 both found defects in the preceding fix pass, which is why each pass was re-reviewed rather than shipped.

### What round 3 caught that I had only half-anticipated

Recording per-message cost turned an OpenClaw/glm group's `runs.cost_usd` from all-NULL into all-zero (that provider reports `usage.cost.total = 0` on every message). `cost_per_quality` then went from NULL — sorted last by `?? Infinity`, never recommended — to **0, sorted first**. `agentdeck apme recommend --budget 3` could answer with whichever model is *worst instrumented*, on the strength of having no price table at all.

`runs.cost_usd` is one REAL column with no room to say why it is zero. Collapsing "free" and "unpriced" at the column is unavoidable; **ranking on the collapsed value is not.** Zero now orders with NULL.

### What round 4 caught

The named comparator I added to fix that broke a rule CLAUDE.md states outright — *comparators are three-way, never subtraction*. Not stylistic: every unpriced key is `Infinity`, so `a - b` is `Infinity - Infinity` = **NaN**, and a NaN comparator leaves the sort order implementation-defined. The old `?? Infinity` had the same flaw, so it is not a regression — but giving the comparison a named home is the moment to honour the rule.

### Verified, not assumed

The reviewer measured rather than reasoned on the two things I flagged as my own biggest doubts:

- **`child.once('exit')` on a `detached` + `unref()` child** — fires (`EXIT fired, code=3`, `childAlive went FALSE at poll 0`). Had it not, my fix would have turned a 20s failure into a 180s hang, strictly worse than the bug it replaced.
- **Explicit `undefined` honours a default parameter** — the 180s ceiling at the call site is intact, not silently undefined.
- **`unpricedLast` edges** — `1e-12` survives; only `0`/`-0`/`NaN` map to Infinity, so no legitimate tiny price is swallowed.

### Every new gate is mutation-verified

Reverting each fix reddens its own test: `114 vs 52222` · `null vs 'stop'` · `2 vs 1` · `10.5 vs 0.5` · `null vs 0` · `0 > 999` · `NaN vs +0` · the liveness wait. One test that stays green under its neighbour's mutation is called out in-file as a contract pin rather than left to read as coverage.

vitest **3491** green · Swift 680 green · CI green on `383235c0`.

### Follow-ups recorded, not silently dropped

`expectedCostUsd` still reports `(totalCost ?? 0) / runs`, so an unpriced model recommended through the *quality* branch reaches API clients as `expectedCostUsd: 0` — the same unsupported claim one field over. It is a wire type with generated Swift/Kotlin mirrors, so it deserves its own review rather than riding out at the end of this chain. Also open: B4 (which of `chat.final` / `session.message` lands second is inferred, not measured), B3 (per-turn model needs a `turns` column), provider ingestion, `messageSeq` gap detection, and the pre-existing 8,000-char prefix collapse.

puritysb:
Follow-ups from this PR are now tracked as issues rather than only in this body: #265 (truncated `session.message` overwriting `chat.final`), #266 (per-turn model written run-level), #267 (`gen_ai.system` has no consumer), #268 (`expectedCostUsd` reports 0 for unpriced models).

PR #282, opened 2026-09-05
PERM fires only on genuine user waits across agents; Claude hold predictor becomes a generated SSOT

puritysb:
## Summary

PERM is a claim that a human is being waited for. This PR re-audits every agent's source for that claim in both daemons and fixes what was wrong, missing, or one-sided.

- **Claude Code** — the observed-session PreToolUse **hold** (device Allow/Deny, 25 s) is the one place PERM is *predicted*, and both daemons' hand-written predictors had drifted into the same defect: they understood only the legacy `Bash(cmd:*)` spelling, while the permission dialog writes `Bash(cmd *)` (67 of the user's 73 Bash rules). Every allowlisted call in an `acceptEdits`/`default` session was held for the full timeout (measured: awaiting→processing exactly 25.0 s, re-held on every batch). The predictor is now a shared SSOT (`shared/src/claude-permission-rules.ts`) emitted into `ClaudePermissionRules.generated.swift` by `pnpm generate-claude-permission-rules` (byte-gated) with behavior pinned by `shared/claude-permission-vectors.json`, replayed by vitest and XCTest. Rules follow the Claude docs (wildcard anywhere, trailing ` *` matches the bare command, compound commands must be fully covered, built-in read-only set, `acceptEdits` filesystem commands, wrapper/env stripping). Learn window 8 s → 15 min.
- **Codex** — no approval signal existed (the rollout carries none), so a session blocked on `approval_policy = "on-request"` read `processing` for 10 minutes. The installer adds `PermissionRequest` and `Interrupt`; Node overlays a display-only PERM, Swift sets `awaiting_permission`, and an interrupted turn closes as `interrupted`.
- **OpenCode** — Swift already surfaced `permission.asked` with an answerable `ocperm:` requestId; the Node daemon dropped it for state (PID-keyed rows). New `bridge/src/hook-opencode-sessions.ts` builds session-keyed rows from the plugin's hooks and lets the PID row for the same cwd yield.
- **OpenClaw** sound as-is; **Kiro/Antigravity** have no permission source and never show PERM (documented).
- **Subagents** — no path mints extra session rows. One latent over-count closed: the timeline-derived fallback kept eviction-exempt dispatch rows while children's stop rows were shed, so after a daemon restart a fan-out read "+8 running" for 6 h. A dispatch older than the parent's own turn close now counts as drained (shared/Kotlin/Swift), and both daemons drop the census on `session_end`.

## Verification

- vitest: 255 files / 3980 pass. macOS XCTest (ClaudePermissionRulesTests, ObservedAskUserQuestionTests, OpenClawToolNoiseTests, CodexConfigInstallerTests): 83 pass. Android `TimelineStoreTest`: 64 pass. `pnpm docs:check`, `pnpm design-system:check`, `pnpm generate-claude-permission-rules --check` green.
- Live: daemon restarted on this build. Synthetic PreToolUse posts — allowlisted `curl` and `mkdir && cp` in `acceptEdits` pass through in ~30 ms; an unlisted `default`-mode command still holds 25 s.
- Not measured live: a real Codex `PermissionRequest` firing (hook installed, path unit-tested).

## Open finding (not fixed here)

`SubagentStart` rarely reaches the daemon (7 days: started 7 vs completed 221; an orchestrator row reads `peak 0 / completed 11`). The hook command differs from `SubagentStop` only by path. Tracked separately.


puritysb:
## Pre-merge review (direct, after two subagent review runs were cut off by the session limit)

Three concerns verified in code and fixed in the follow-up commit:

1. **Learner race with the widened window** — `noteToolEnd` matched any PostToolUse of the same *tool name*; with a 15-minute window an unrelated Bash could teach the held signature when the Notification was dropped. Now keyed by the held call's `tool_use_id` in both daemons (name fallback only for releases without an id).
2. **Codex PERM cleared by transcript recency** — the Claude ESC heuristic would drop the Codex overlay if the rollout was appended during the wait. Codex overlays are `sticky`; hook clears + TTL own their lifetime.
3. **Legacy `:*` semantics** — docs say alias of ` *`, older builds read a raw prefix. Both readings accepted (over-matching = missed hold, the safe direction). Vectors updated.

Also checked: child hooks return before the Codex overlay block (a `codex_subagent_*` cannot clear it); hook-derived OpenCode rows (`pid: 0`) follow the existing Codex pattern; `_interrupt` stamping precedes the collector call.

PR #284, opened 2026-09-06
fix(usage): recover Claude authorization and explain missing quota

puritysb:
## What changed

Claude quota disappeared while the daemon was healthy because its OAuth credential had expired. The Node daemon now requests bounded recovery through the user's Claude CLI, re-reads the credential, and retries the usage request. The app displays the relayed authorization failure separately from session connectivity and clears it when fresh data arrives.

The recovery turn uses Haiku with tools, hooks, MCP and session persistence disabled, a 25-second timeout, and a requested $0.01 budget. It may consume a small amount of subscription quota. Persistent cooldown is 30 minutes, increasing to six hours after three attempts; `AGENTDECK_CLAUDE_USAGE_RECOVERY=0` disables it. AgentDeck does not rotate or write Claude refresh tokens.

API retries now use failure timestamps and separate `Retry-After` metadata instead of making old cache timestamps look fresh. Review also fixed a transient-401 latch: an unchanged unexpired credential can be retried after backoff, even when CLI recovery is unavailable. Legacy fresh usage frames clear prior authorization errors. Relative credential directories are resolved before changing the child working directory; custom macOS namespaces are skipped because the usage reader observes the default Keychain service.

## Validation

- [x] Isolated branch: `pnpm build`, `pnpm -r exec tsc --noEmit`, and `pnpm docs:check` passed.
- [x] `pnpm test`: 257 files, 4,007 passed, one skipped because the optional prebuilt FM helper is absent.
- [x] Follow-up credential-store/API regression tests: 19 passed; bridge build and documentation check passed.
- [x] Final commit `ea7ceb5f`: all nine PR CI checks passed.
- [x] Focused macOS XCTest: 21 passed, including wire-state recovery and ownership changes.
- [x] Actual bounded Claude invocation exited normally: one Haiku turn, reported cost $0.000644. Expired/rejected credentials, failed renewal, retry persistence, 429 and concurrency use isolated regression tests; the live account's token was not forcibly expired.
- [x] Documentation and feature matrix updated; no generated protocol shape changed.
- [x] No credentials, private diagnostics or unrelated APME work included.
- [x] [Distribution-signed iOS/macOS archive validation](https://github.com/puritysb/AgentDeck/actions/runs/34006409630) passed with `upload=false`; both upload steps and release were skipped. This run validated `55fedd19`; the final `ea7ceb5f` has identical Apple sources and release workflow. The local execution app remains development-signed Release.

## Platform and release impact

- [x] Swift adds status handling only; no subprocess, interpreter, credential acquisition or companion-install prompt.
- [x] Existing optional `tokenStatus` is now explicitly emitted for `unknown` too, so stale errors can be retracted. Protocol remains compatible.
- No App Store upload, submission, release or merge is part of this PR creation.

## AI assistance

Implemented and reviewed with Codex. The review traced the Node acquisition/retry path, child-process boundaries, Swift event merging and standalone ownership transition; the two recovery-state issues and the credential-store boundary were fixed and covered by regression tests. No independent human review is claimed.


puritysb (commented):
Codex self-review of `55fedd19675001cf3d16f6e6db82f0d4447b88fd` completed. This is an assistant review, not an independent human approval.

Two findings were corrected before this commit:

1. A transient 401 could permanently block an unchanged, unexpired token when CLI recovery was unavailable or unnecessary. The API now retries after backoff; a later response clears the rejection marker. Regression cases cover recovery enabled and disabled.
2. A legacy fresh usage frame that omits `usageStale: false` could leave an old authorization error retained for the next outage. The Swift reducer now clears it on live quota data too, with a wire-level regression test.

Reviewed recovery process arguments, timeout and persisted retry limits; credential re-reading and API/cache freshness; Swift status propagation and ownership transition; and App Store boundaries. No additional actionable P1/P2 findings identified in this review.

The isolated branch passed build, TypeScript and documentation checks, 4,007 Node tests (one optional prebuilt-helper check skipped), and 21 focused macOS tests. PR CI and the distribution-signed `upload=false` archive run are being checked separately. Unrelated APME work is excluded.

puritysb (commented):
Final Codex self-review of `ea7ceb5f37828c316173f41b5d2eec735c757131` completed. No additional actionable P1/P2 findings identified; this is not an independent human approval.

In addition to the two state-recovery fixes recorded in the earlier review, the environment pass found that a relative `CLAUDE_CONFIG_DIR` could resolve against the recovery child's temporary cwd, and a custom macOS directory could select a different Keychain namespace from the usage reader. The follow-up commit resolves relative directories before launch and skips mismatched macOS namespaces. Regression tests also assert removal of inherited credential/provider/session overrides without modifying the parent environment.

All nine CI checks pass on the final commit. The separate distribution-signed iOS and macOS verifier run passed with uploads disabled: https://github.com/puritysb/AgentDeck/actions/runs/34006409630 . It ran at `55fedd19`; Apple sources and the release workflow are unchanged in the final commit. Focused macOS tests: 21 passed. Follow-up environment/API tests: 19 passed.

The reviewed Release app and Node daemon are running locally, with live Claude quota present and `tokenStatus: valid`. The isolated PR checkout is clean and excludes the unrelated APME work.

puritysb:
Adversarial review before merge (two independent passes: correctness on the Node bridge, contract/regression on Swift + project invariants). Fixes pushed as `8909ecdd` + `f24330b9`.

**Blocker, confirmed and fixed — the standalone App Store build showed a permanent quota failure.**
`UsageAPIClient.directOAuthUsageSupported == false` means the sandboxed daemon's `tokenStatus` can never leave `unknown` (the guard returns before every write), `usageStale` is always true with no relay, and `effectiveOauthConnected()` returns true whenever any `claude-code` session is cached. `claudeUsageIssue`'s `default:` branch fired on that trio, so every ordinary standalone user saw "Usage temporarily unavailable" replacing the Claude model catalog, "Hooks on", and the USAGE / RATE LIMITS headers — the opposite of the "reads as feature-complete, never as broken" invariant in CLAUDE.md. The doc comment on the property and the feature-matrix line ("displays the **relayed** failure reason") were false for the same reason.

Now only an explicit `expired` is a failure claim. `missing` is not one either: an API-key or off-harness install has no OAuth credential by design, and `usageStale` means "no numbers", never "authorization failed". Reverting that one line fails 5 assertions in the new tests.

**Also fixed**
- Recovery was unreachable for credentials with no numeric `expiresAt` (`CLAUDE_CODE_OAUTH_TOKEN`, `claude setup-token`): both halves of the branch were gated on it. Measured probe before the fix — three 401 polls across 20 minutes, `recover()` called 0 times, `tokenStatus` latched `expired`. The 401 memory now opens the branch on its own (both lines, 411 and 421 — either alone is a no-op).
- `execFile('claude')` consults neither PATHEXT nor a shell, so recovery was ENOENT on every Windows install and on any daemon whose baked LaunchAgent PATH lacks the CLI — visible only as a recurring generic failure line. The binary is resolved off PATH once; a Windows shell shim is declined with one explicit line rather than shell-quoting a JSON argument through cmd.exe (`AGENTDECK_CLAUDE_CLI` overrides).
- The cooldown file is written tmp+rename: a torn write reads back as "no cooldown", which is the one failure mode that spends quota.
- Display precedence: "Not connected" outranks a usage sentence, and the rail appends the reason to the model catalog instead of replacing it.
- The new `fileCacheExpired` guards were uncovered (reverting them left all 36 usage tests green) — they are the migration path for a cache the old 429 code stamped in the future, so they are pinned now.

**Verified clean by the correctness pass, measured rather than reasoned:** the 25 s timeout fires at 25,010 ms with SIGKILL and no surviving descendants; one real invocation costs $0.000654 and creates no AgentDeck session or timeline row; the persisted cooldown is genuinely read on the deciding path and written before the spawn (both mutation-red), capping recovery at ≤1 spawn per 30 min per daemon even under client-triggered `query_usage`; and no branch records a failed fetch as fresh. 11 mutations + 3 live probes; exactly three production lines stayed green and all three are now either tested or filed.

Deferred to #286: the Anthropic judge leg's missing truncation guard, Node-only `reasoningEffort`, the `AGENTDECK_DATA_DIR` inconsistency shared with `usage-api.ts`, and the orphaned-child-on-SIGKILL observation.

PR #285, opened 2026-09-06
fix(apme): reject incomplete verdicts across daemons and preserve serving evidence

puritysb:
When an MLX or OpenAI-compatible judge reaches its output limit, its content can still look like valid JSON. Both daemons now reject that response before it becomes a verdict. Swift uses one response gate for both adapters, Node also rejects whitespace-only MLX content, and both suites replay the same response vectors.

Deployment verification also exposed a stale hub port label after Swift-to-Node handoff; the dashboard now prefers the actual connection URL over a retained partial state field.

This PR preserves the previously uncommitted APME serving work: optional Node `reasoningEffort`, real SQLite persistence/recovery tests, and three private replay tools. Replay profiles record source/build/input/script hashes and server identification, refuse changed conditions and unsafe output paths, and preserve interrupted attempts. Historical MLX/APC/Ollama results and their limitations are retained; completion rate is not presented as judge accuracy. No production database replay or backend change is performed.

Depends on #284. The base is `codex/claude-usage-recovery`, so the diff shows only this follow-up. Merge #284 first, then retarget this PR to master. General CI (9 jobs) passed on the final head e6de29b2 using a temporary master comparison, because repository CI only triggers for master-targeting PRs. No commits were rebased or changed when restoring the review base.

Validation:
- Full Vitest: 259 files, 4,051 passed, 1 skipped (optional prebuilt Foundation Models helper absent); subsequent interrupted-attempt guard: 12 script tests passed. Final persistence/default-effort review tests: 29 passed.
- macOS judge XCTest: 20 passed, including shared vectors.
- Final general CI: all 9 jobs passed, including Windows Node 22/24/26, macOS XCTest and iOS build. Node CI: 4,054 passed, 2 skipped.
- Workspace build, typecheck and docs check passed.
- Synthetic capture through the built task runner: four axes, task/run links and persistence-before-event passed; identical replay resumes, changed provenance is rejected. No live model inference was used for this check.
- Release macOS build passed. Local archive verifier rejects only the development signing identity. Distribution-signed iOS/macOS archive validation passed on final head e6de29b2 with upload=false: https://github.com/puritysb/AgentDeck/actions/runs/34008162612. No store upload or release is requested.

Review: Codex self-review, not independent human approval. Checked Swift/Node response parity, fallback boundaries, private evidence isolation, and checkpoint overwrite behavior. Human-labelled judge accuracy comparison remains a documented follow-up.


puritysb (commented):
Codex self-review of e6de29b2 (not independent human approval).

The APME review range starts at ea7ceb5f, the head of dependency #284. Checked:
- MLX/OpenAI response rejection happens before parsing and persistence in both daemons; a valid JSON body with finish_reason=length is rejected. Missing finish_reason remains supported for legacy servers.
- Shared response fixtures run in Vitest and XCTest. Real SQLite failure/recovery tests preserve four axes, the run association, persistence-before-event, and manual outcome overrides.
- reasoningEffort is optional and remains Node-only; omission preserves the provider default. The running judge is still MLX.
- Replay output paths cannot escape through case/profile names or output symlinks. Source/build/input/script hashes fix provenance, redirects are blocked, and exclusive start records preserve interrupted attempts. Synthetic replay and resume/mismatch checks passed without live model inference.
- App Store Swift paths remain network-only. Final local Release app is development-signed with sandbox enabled and get-task-allow removed; distribution archive verification runs separately with upload=false.
- Deployment exposed a stale :9121 hub label after connection to :9120. The final app now displays the actual :9120 connection and live Claude usage.

No unresolved actionable finding in the reviewed changes. General CI and distribution archive checks passed on the preceding revisions; the same checks are running on the final port-display commit. Judge accuracy against frozen human-labelled cases remains a separate, explicitly unfinished follow-up; serving completion counts are not accuracy claims.

puritysb:
Final verification completed for e6de29b2:

- All 9 general CI jobs passed (Node: 4,054 passed / 2 skipped; Windows Node 22/24/26; macOS XCTest; iOS build; submission/design/ESP32 checks).
- Distribution-signed iOS and macOS archive/export verification passed with upload=false: https://github.com/puritysb/AgentDeck/actions/runs/34008162612. No App Store upload or release was performed.
- Restored the PR base to codex/claude-usage-recovery so the diff contains the 18-file follow-up only. Merge dependency #284 before retargeting to master. Head and source tree did not change during the CI base switch.
- Local Release app and CLI are running from this committed source. Live checks show hub :9120, Claude usage, connected devices, and APME MLX ready. The normal pnpm and Homebrew CLI entrypoints now resolve to this worktree; running and on-disk daemon build hashes match.
- Local app uses the current team's development identity with App Sandbox enabled and get-task-allow removed. Formal distribution signing was validated separately in CI.

The earlier comment is a Codex self-review, not independent human approval. The remaining APME quality work is a frozen dataset with human-confirmed reference judgements, as documented in docs/apme.md; the preserved serving experiments do not claim to establish judge accuracy.

puritysb:
Adversarial review before merge; fixes pushed as `7e8028fc` (plus the #284 fixes merged up).

**The port fix had no gate.** `daemonPortText` was a `private var` on a SwiftUI `View`, unreachable from the test target, and `ProviderRailEvaluatorTests` gained no port case — CI could not fail on that commit. Extracted to `HubPortRules.portText` with the four cases pinned (connected URL beats a retained `daemonPort`, no URL, neither, and `0` meaning never populated); swapping the order back fails the new test. The commit's premise is in fact stronger than its comment claimed: `daemonPort` is emitted by the Swift daemon only and is absent from `shared/src/protocol.ts`, so after a Swift→Node handoff the retained value is guaranteed stale.

**Truncation rejection is now visible.** Rejecting `finish_reason: length` sends the default chain to the Foundation Models floor — a measurably weaker judge with a 4× smaller window — and the only trace was a DEBUG line reading "mlx unavailable", which is a different fact from "the answer was cut off". It prints in the normal log now, so the rate is measurable before it matters. Bounding measurement on this machine's store (30 days): verdict bodies p50 697 / p99 1,025 chars (~335 tokens against an 800-token cap) and 759/759 verdicts from MLX, 0 from FM — so the cap is not being reached here today.

**Verified, not assumed:** both suites really do replay `shared/apme-judge-response-vectors.json` — deleting the Swift `finish_reason` gate fails `testChatResponseMatchesSharedVectors` under a real `xcodebuild test`, and deleting both Node throws fails 7 tests — so a vector added to that file would fail a daemon that ignores it. The replay scripts' guards hold: the private root refuses the repo, `~/.agentdeck` and `AGENTDECK_DATA_DIR` including symlink aliases (realpath-based, evaluated before the scripts repoint the env), outputs refuse symlinks, `fetch` is replaced with one that admits only 127.0.0.1:8800/11434 with `redirect: 'error'`, and the production SQLite is never opened.

Filed as #286 rather than fixed here: the Anthropic `api` leg has the identical truncation defect and the two daemons' caps differ 8× (1,024 vs 8,192); `reasoningEffort` is Node-only, so identical settings that make Ollama work for Node now make Swift hard-fail on this PR's gate; and `choices` as a non-array is a Node/Swift divergence the vector file does not pin.

Issue #299, opened 2026-09-07
APME judge penalty: deferred findings from #298's five review rounds


puritysb:
Follow-up from the five adversarial review rounds on #298. Everything here was
**found and verified but deliberately not fixed there** — either out of that
PR's scope, or a decision that wants its own discussion. Each item names the
file and the input that reproduces it, so none of this needs re-deriving.

## 1. The suppression mechanism is a deletion candidate

**22 production references across both daemons**, defending against a condition
**never once observed**: a judge server that rejects `repetition_penalty`. The
server this was all measured against silently ignores unknown fields (CLAUDE.md
records this — a made-up parameter returns HTTP 200 exactly like a real one).

It produced a HIGH or MEDIUM finding in **four consecutive review rounds**:

| round | finding |
|---|---|
| 2 | penalty extended to the `openai` leg (silently altering paid sampling) |
| 3 | Swift `send` collapsed all failures → a repetition cut permanently disabled the penalty |
| 4 | suppression recorded before the retry proved anything; no `> 1` gate on Swift; `classify` read all 4xx as a refusal |
| 5 | attribution blamed the expensive field on ambiguous evidence; Node/Swift disagreed on what "the retry succeeded" means |

This is the repo's own documented delete-the-mechanism criterion (see the #286
brace-topology exemption, removed on exactly this basis: repeated defects, no
measured beneficiary).

**The alternative keeps the protection and removes the state**: retry once
without the field per request, and don't remember. Cost on a hypothetical
strict server is one extra immediate 400 per judge call instead of one per
process — judge calls already run 10–60 s. The win is not fewer lines; it is
that every remaining undiscovered bug in this area changes from *permanent,
silent, process-wide* to *one wasted request*. All four findings above were
HIGH/MEDIUM **because** of that permanence.

Counter-argument, stated fairly: "never observed" is not "cannot happen" —
`apme.judge.endpoint` may point at vLLM, LM Studio or OpenRouter, and some may
genuinely 400. That is a plausible guess, not a measurement, and this PR's whole
discipline is claiming only what was measured.

## 2. Cross-daemon divergences (pre-existing, all verified in round 5)

Both daemons may hold port 9120 and read the same `settings.json`, so each of
these means the same user config behaves differently depending on which daemon
answered, with nothing in either log saying so.

- **`ApmeJudgeMlx.swift:58`** — the MLX leg hardcodes
[...]

Issue #286, opened 2026-09-06
apme(judge): the truncation guard covers two of four legs, and reasoningEffort is Node-only


puritysb:
Follow-ups deliberately left out of #285, which guarded the MLX and OpenAI-compatible legs only. None of these is a regression from that PR; each is a gap the PR made visible.

### 1. The Anthropic `api` judge leg has no truncation guard, and the two daemons truncate at different budgets

`ApmeJudgeApi.swift:55` sends `max_tokens: 1024` and never inspects `stop_reason`; Node `bridge/src/apme/runner.ts` sends `max_tokens: 8192` and checks only `stop_reason === 'refusal'`, not `'max_tokens'`. So the widest unguarded path is also the one where the two daemons disagree 8×: the same task, judged on the same settings, can be cut at different points and both bodies are accepted as verdicts.

- Reject `stop_reason === 'max_tokens'` in both API legs.
- Align the two `max_tokens` values, or record why they differ.

### 2. `apme.judge.reasoningEffort` is Node-only, and the new gate turns its absence into a hard failure

Zero hits for `reasoningEffort` under `apple/`. Both daemons read the same `settings.json` and call the same user-configured endpoint. A user sets `reasoningEffort: 'none'` to stop an Ollama model emitting thinking tokens: Node obeys, Swift does not, so Swift's request emits thinking, exceeds `max_tokens: 1024`, returns `finish_reason: "length"` — and #285's gate now rejects it. Net effect: the Swift daemon cannot judge with a model the Node daemon judges fine, from identical settings.

Mirror the field into `ApmeJudgeOpenAI` (one key in the request body), or raise Swift's cap.

### 3. `finish_reason: length` discards a complete, parseable verdict and silently reroutes the default chain

The throw happens before the content is read, so a verdict that merely ended at the cap is dropped. On the default chain (`mlx` + `fallbackToFoundationModels`) it lands on the Foundation Models floor — 0.580 on the judge-fidelity rubric against 0.86–1.00 for the MLX tier, with a 4× smaller window. #285 makes the switch visible in the normal log; it does not decide the policy.

[...]

PR #301, opened 2026-09-09
fix(apme): a run opens on a span that can record, and one settings file gets one verdict (#299, #300)

### LINKED ISSUE #300: OpenClaw: a session key still opens an empty APME run when every span it carries is discarded
Measured on the live daemon 2026-09-08, while verifying #296's watchdog fix.

## What was measured

Of the 6 OpenClaw runs created since the daemon restart (2026-09-07 22:15 KST),
**three hold zero turns, zero sample_events and zero tasks**, and were closed
only by the 30-minute orphan reaper:

```
proj                        started              dur_s  turns  events  tasks
openclaw (connection)       2026-09-07 22:15:22   1804    0      0       0
agent:main:cron:41b100c9    2026-09-07 23:50:00   1046    1     18       1   ← correct
agent:main:cron:41b100c9    2026-09-07 23:55:31   1046    1     10       1   ← correct
agent:main:cron:6742c6d3    2026-09-08 00:00:00   1107    1     10       1   ← correct
agent:main:main             2026-09-08 00:01:34   1822    0      0       0
agent:main:main:heartbeat   2026-09-08 00:01:48   1808    0      0       0
```

The `openclaw` connection-scoped row is the **known-open** item already recorded
in CLAUDE.md. The other two are not: CLAUDE.md states the lazy-open change fixed
exactly this, naming `agent:main:main:heartbeat` and "0 turns / 0 steps / 0
events" as the symptom it removed.

## Why the fix is one layer too shallow

`OpenClawAdapter.ingestApmeSpans` only declines to open a run when the span list
is **empty**:

```ts
if (!apme || spans.length === 0) return;
const target = sessionKey ? this.apmeSessionIdFor(sessionKey) ?? … ;  // ← opens the run
```

But `openclawSessionMessageToSpans` builds spans for an assistant
`session.message` regardless of whether a turn is open — a `session_meta` span
when the message carries `model`/`provider`/`usage`, plus a `turn_response` when
it has text. So the array is non-empty, the run opens, and then
`Collector.ingestSpan` discards every one of them:

- `turn_response` → `setTurnResponse` → `if (!turnId) return`
- `session_meta` → `updateTurnIdentity` / `addUsageIncrement`, both no-ops with
  no open turn

Result: a run row with nothing in it, held open until the 30-minute reaper. The
guard's predicate is "are there spans?" when the honest predicate is **"is there
a span that will actually be recorded?"** — i.e. the same shape as the defect the
lazy open was introduced to fix, moved down one level.

`agent:main:main` doing this is the more interesting half: that is the user's own
chat key, not a heartbeat.

## Not proposing the fix here

The obvious options (have `ingestSpan` report whether it recorded anything; or
open the run only on a span kind that can stand alone, e.g. `turn_start`) both
touch the collector's contract, which is shared by every adapter. Worth deciding
deliberately rather than inside an unrelated PR.

Related context in CLAUDE.md: the OpenClaw session-key scoping section, and the
"Still open" note about the connection-scoped fallback run.

puritysb:
Closes the two open APME issues. Everything below was measured against the live daemon or driven through the real code; nothing was re-derived from the issue text.

## #300 — the empty OpenClaw runs

`ingestApmeSpans` asked *are there spans?* when the honest question is *is there a span that will actually be recorded?*. An assistant `session.message` always builds a `session_meta` (it carries model/provider/usage) and, with text, a `turn_response` — so the array is never empty. The run opened; the collector then discarded both for want of an open turn.

Measured on `~/.agentdeck/apme.sqlite`, week to 2026-09-09:

| | runs |
|---|---|
| OpenClaw runs | 113 |
| …with zero `sample_events` | 87 |
| …with zero `steps` as well | 82 (25 per-key, 57 connection-scoped) |

All 82 were already tagged `_empty`, so the dashboard filtered them — the cost was a held reaper slot per run and a polluted denominator for any OpenClaw measurement, not a visible wrong number.

`ApmeCollector.spanCanOpenRun` is the predicate, deliberately adjacent to `ingestSpan` (it is a claim *about* that dispatch) and exhaustive, so a new `TelemetrySpanKind` cannot compile until it is classified. The four `true` kinds are exactly those reaching `ingestHook`, which inserts a `steps` row unconditionally once the run exists.

The connection-scoped fallback — CLAUDE.md's "Still open" item — follows the same rule. Its id is still minted at connect so `hasDirectApmeIngestion` (which gates the bridge's timeline→span conversion) answers unchanged; only `openRun` is deferred. Wiring that supplies no opener still reads as "already open", which preserves the pre-split fallback for a key that can never get its own run.

## #299 — the same settings file, two verdicts

Both daemons may hold 9120 and both read one `settings.json`, so each of these means the same configuration scores differently depending on which daemon answered, with nothing in either log saying so. All seven items were verified still-present in code before touching anything.

- **MLX judge URL** resolves `apme.judge.endpoint` then the `llm.mlx` pin, as Node does. Swift read that pin for the *model* and hardcoded the URL to loopback, so `{"llm":{"mlx":{"endpoint":"http://192.168.1.5:8800"}}}` had Node judging against the LAN server while Swift posted to loopback, failed, and dropped silently to the FM floor (0.580 vs 0.86–1.00 on the judge-fidelity rubric).
- **MLX judge timeout 60 → 90 s**, matching `runner.ts` — and 68.7 s under local load is where the 90 came from, on this exact prompt shape.
- **MLX → FM fallback logs at `info`.** Node's comment states the requirement ("or its rate is unmeasurable"); behind Swift's debug flag, the cut rate the 1.05 repetition-penalty default is calibrated against could not be measured on that daemon at all.
- **`guard code == 200` → `200...299`**, mirroring `resp.ok`. Round 5 widened the MLX leg and left its sibling. The BODY gate is untouched — a 201 with a cut body is still not a verdict, and that is asserted.
- **The MLX ladder writes nothing off when the prompt was compacted after a drop.** That success differs from the failure in two ways, and the other one — no longer overflowing the context — explains it by itself. The field/field asymmetry is about choosing between two fields; it does not reach a case where a field may not have been the cause at all.
- `if (compacted === body) break;` now has a gate.

`ApmeJudgeOpenAI` gets the injectable transport `ApmeJudgeMlx` already has. The status gate is one line inside the call, not a pure function anywhere, and #298's round 5 recorded what happens when "this path has no tests" is answered with one more pure seam: the hole reopens a layer up.

**Left open on #299** (unchanged, and stated there): Swift has no `response_format` ladder at all (`grep -rc response_format apple/` = 0), and the two `task_category` classifiers differ in backend, `max_tokens`, timeout and fallback — which selects a different rubric, so it is a score difference. Item 1 (delete the suppression mechanism) is a design decision the issue explicitly reserves for its own discussion and is not touched here.

## Release and doc hygiene

- **`## Unreleased` in CHANGELOG.md.** #294–#298 plus the deck, ESP32-rename and CLI-restart fixes had no entries — the documented recurring failure (the 1.2.1 cut had to reconstruct #282–#289 from a commit log). The heading cannot be matched by any tag lookup, which is pinned by a test; a cut renames it.
- `docs/apme.md` loses two stale claims: `reasoningEffort` has not been Node-only since #286, and the JSON-mode ladder *is* Node-only, which the doc did not say.
- `experiments/collaboration/README.md` still opened with "a local-only experiment on `codex/dashboard-collaboration`" for work that shipped in #291 / Apple 1.2.1. It now carries a status header; the trial record itself is unchanged.

## Verification

- vitest 267 files / 4,194 passed, 1 skipped
- macOS XCTest 779 passed / 2 skipped / 0 failed
- `tsc --noEmit`, `docs:check` (116 files), `verify-version`, `design-system:check`, `verify-tokens-sync` all clean
- **Every new test was mutation-checked.** Forcing `spanCanOpenRun` to `true` turns 3 red; removing the compaction/attribution guards turns 2 red; restoring the two Swift hardcodes turns 2 red.

Closes #300

Issue 299 stays open — the two items above (`response_format` on Swift, classifier convergence) and its item 1 are not addressed here.


puritysb:
## Self-review pass — one real defect found and fixed

Reviewed the diff adversarially before merging rather than merging on green CI. Two changes, one of them a genuine bug I introduced.

**1. The lazily-opened fallback run lost its model id.** `openclawRunForSessionKey` seeds `modelId: gatewayModelName` at open, and its comment says exactly why: *"a run opened AFTER `model_info` fired would otherwise persist `model_id=NULL` — the defect that per-run `updateModel` was added to fix, reintroduced by the lazy open."* The connection-scoped run got its model from the `model_info` broadcast at line 4917, which now no-ops because `updateModel` guards on `sessionToRun.has(...)` and the run may not exist yet. So making the fallback lazy reintroduced that defect for it. The deferred opener now seeds `modelId` the same way, read at open time so it carries what the catalog last reported.

**2. Log prefix.** `DaemonLogger.info` takes no category, and the repo's convention is `Component: message`, not a bracketed tag. It is now `APME judge: mlx produced no verdict — falling back to foundationModels` — byte-comparable with the Node daemon's line, which is the point of making it visible at all.

### Checked and found clean

- The other two uses of `openclawApmeSessionId`. `/apme/task/close` defaults to it: `closeTaskExternal` on a missing run returns false and the route answers **404**, so an unopened fallback cannot be reported as a closed task. `closeRun` on disconnect is a no-op for a run that was never opened.
- `hasDirectApmeIngestion()` is still `apmeSessionId != null && getApme() != null` — the id is minted at connect exactly as before, so the timeline→span gate in `index.ts` is untouched.
- **Endpoint parity is exact.** Swift's `stripChatSuffix` strips `/v1/chat/completions` then `/chat/completions`, anchored at the end; `shared/src/llm-settings.ts` does the same with anchored regexes, and both then append `/chat/completions`. Every input form resolves to the same URL on both daemons.
- `spanCanOpenRun` against what OpenClaw actually emits (`turn_start`, `turn_response`, `turn_end`, `tool_call`, `tool_result`, `task_boundary`, `session_meta`, `agent_error` — no `raw_step`): the only kinds that open a run are ones whose step rows survive `closeRun`'s `_empty` filter.
- Widening the status gate does not widen the body gate — a 201 carrying a `finish_reason: "length"` body is still refused, and that is asserted.
- `compactedAfterDrop` only suppresses attribution when the compaction happened **after** a drop. Compaction-then-drop still attributes to the drop, correctly: that success differs from the preceding failure only by the field.

Re-verified after both fixes: vitest 267 files / 4,194 passed, macOS XCTest 779 passed / 2 skipped, tsc clean.

Issue #303, opened 2026-09-09
[Bug]: Connected: AgentDeck [DISCONNECTED]


jurri:
## Maintainer status — 2026-09-15

Still awaiting a reproducible plugin/daemon connection trace. PR #304 fixes checkout instructions; #329 and #330 address Windows hooks and observer state. None proves that the Stream Deck OFFLINE symptom here is resolved. Next check after the forthcoming npm patch: record installed CLI/plugin versions, daemon/native diagnostic results, and sanitized Stream Deck connection logs; distinguish WebSocket connection failure from an idle/disconnected agent state.

---

### Affected surface

CLI or daemon

### Version

1.2.0.0

### Environment

Windows 11
ThinkPad P16s Gen2
npm/x 11.13.0
node v24.16.0

### Reproduction steps

`agentdeck daemon install` worked fine and I can use it on Stream-Deck.

after closing and `agentdeck daemon start` I get ever 'Offline'-display on Stream-Deck. 

on `agentdeck dashboard` I can see limit, Models etc. 
and this:  Connected: AgentDeck [DISCONNECTED]

stop, start, restart of agentdeck and Stream-Deck doesn't help.

### Expected behavior

I see something on Stream-Deck

### Actual behavior

I see only 'OFFLINE'

### Sanitized logs or screenshots

```shell

```

### Checklist

- [x] I searched existing issues for the same problem.
- [x] I removed secrets and private information from this report.

puritysb:
Thanks for the report, and for separating what works from what doesn't — that narrows it a lot already.

**First, one line you can stop worrying about.** `Connected: AgentDeck [DISCONNECTED]` is not an error and does not describe the dashboard's own connection. It is the first timeline row the TUI writes after it attaches, and it prints the *agent session* state — `DISCONNECTED` there means "no Claude Code / Codex / OpenCode session is currently being observed", which is the correct state when none is running. The fact that you see limits and models at all means the dashboard found and connected to the daemon successfully.

**So the question is narrower than it looks:** the dashboard reaches the daemon, and the Stream Deck plugin does not. `OFFLINE` on the keys means exactly one thing — the plugin's WebSocket to the daemon is down. Both read the same file to find it (`%USERPROFILE%\.agentdeck\daemon.json`), so the interesting part is what differs between the two processes.

Could you grab these five? Any one of them may be enough.

1. **`agentdeck status`** — names the port the daemon is actually on and whether more than one is registered.
[...]

PR #304, opened 2026-09-09
docs(windows): run the checkout CLI without a global link

jurri:
The Windows checkout guide used `pnpm link --global`, which the contributor's pnpm rejects. It now runs `node bridge/dist/cli.js daemon install` directly and consistently uses the built CLI for checkout commands, so setup no longer depends on a global link.

The guide distinguishes daemon autostart from putting `agentdeck` on PATH, explains that the Scheduled Task references the checkout, and retains `npx @agentdeck/setup` for normal package installations. This does not claim to fix the separate Stream Deck connectivity report #303.

The contributor's original commit is preserved; a maintainer follow-up updates the guide against current master and completes the direct-invocation instructions.

Validation: `pnpm docs:check` and `pnpm design-system:check` pass. Clean-source design lint is 89, matching the base. All CI checks passed before merge, including Windows Node 22/24/26. Merged with the original contributor commit preserved.

### DIFF
--- a/docs/windows.md
+++ b/docs/windows.md
@@ -29,7 +29,7 @@ pnpm test               # optional: run the Vitest suite
 node hooks/dist/install.js
 
 # Link the CLI + Stream Deck plugin
-cd bridge; pnpm link --global; cd ..
+node bridge/dist/cli.js daemon install
 cd plugin; streamdeck link bound.serendipity.agentdeck.sdPlugin; cd ..   # then restart the Stream Deck app
 ```
 


puritysb:
Thank you — you found a real defect, and it is worse than the docs line. But I would rather not merge this diff as it stands, and I want to be precise about why.

## What I could and could not reproduce

On pnpm **11.5.2** here:

```
$ pnpm link --help
Usage: pnpm link <dir>
```

No `--global`, and no `-g` — so the command in the docs is **undocumented**. But passing it anyway does *not* fail at parse time on this version; it gets as far as complaining about my global bin directory. Your `unexpected argument '--global'` is a clap-style rejection from pnpm's Rust front-end, which means your pnpm parses it more strictly than mine does.

So I can confirm the flag is undocumented and version-dependent, and I cannot reproduce the exact rejection. Could you post `pnpm --version` and the verbatim error? That is the one fact that decides what the line should say.

I am deliberately not guessing a replacement flag either: pnpm's documented alternatives for installing a local package globally resolve `workspace:` dependencies, and `@agentdeck/bridge` depends on `@agentdeck/shared` through the workspace — which is why the plain symlink form was used here in the first place. Naming a substitute I have not run on a checkout would just move the problem.

## Why not this diff

`pnpm link --global` and `agentdeck daemon install` do two unrelated things: the first puts `agentdeck` on your PATH, the second registers the autostart Scheduled Task and installs the agent hooks. Replacing one with the other drops the linking step from the Windows guide entirely — which I think your own note is pointing at ("if u don't close the terminal :D"). The doc needs both, and it already has `daemon install` further down.

## What I did fix

`scripts/install.sh` — the `pnpm setup` path — was doing this:

```sh
pnpm link --global 2>/dev/null || {
  warn "pnpm link failed — you may need to link manually"
}
ok "agentdeck CLI linked"      # ← printed either way
```

It discarded pnpm's error, then reported the link as done regardless. So on your pnpm the installer claims a link it never made, and whatever `agentdeck` you type afterwards is either missing or some unrelated global install. That is now reported by its outcome, prints what pnpm actually said, and names the two paths that never depend on the link at all:

- `node bridge/dist/cli.js <args>` — every `agentdeck <args>` in the docs works in this form on a checkout, which is also how you validated your own change
- `npx @agentdeck/setup` — for a normal install that is not a source checkout

`docs/windows.md`, `docs/linux.md` and `docs/install.md` now carry that caveat and link back to your report. Pushed as `79beab57`, credited to you.

Separately: your #303 is a different problem — the Stream Deck plugin cannot reach a daemon the TUI dashboard reaches fine — and I have left a set of commands there that should pin it down. If your CLI was never actually linked, that is worth ruling out first: run the same checks with `node bridge/dist/cli.js status` so we know we are asking the daemon you think we are.

Happy to reopen this shape once we know your pnpm version — if `--global` is gone for good in a version people will have, the docs need a real replacement rather than a caveat, and I would rather land that with your error message attached to it.

puritysb:
Thank you for contributing the direct `node bridge/dist/cli.js daemon install` path. I am revisiting this PR to get that useful workaround merged. The remaining documentation issue is that daemon registration and putting `agentdeck` on PATH are separate operations.

With maintainer edits enabled, I will add a small follow-up to make the checkout instructions consistently use your direct Node invocation, remove their dependency on the unsupported global-link step, and distinguish checkout usage from `npx @agentdeck/setup`. This will preserve your original commit and credit. It will not claim to fix the Stream Deck connectivity issue in #303.

puritysb (approved):
The direct Node invocation is now used consistently throughout the checkout guide, with autostart and PATH registration clearly separated. The original contribution is preserved. Local documentation/catalog checks pass, and all current CI checks, including Windows Node 22/24/26, are green. Thank you for identifying the broken setup step.

Issue #287, opened 2026-09-06
Experiment: macOS collaboration lens and IPS10 work cards — observation and handoff


puritysb:
## Goal and status

Track the locally deployed macOS Dashboard collaboration lens and IPS10 work-card experiment, so another agent/session can resume without the original conversation.

The direction is **preserve the terrarium, improve work/coordination legibility, and adapt presentation per surface**. Evolving harness capabilities should enrich shared evidence rather than force us to reinvent a harness-specific UI or orchestration engine. Unknown relationships are not inferred from project membership or prose.

**Implemented and locally deployed; observation/integration remain open. No public release or merge is requested by this issue.**

## Start here

- [Canonical trial and handoff document](https://github.com/puritysb/AgentDeck/blob/codex/dashboard-collaboration/experiments/collaboration/README.md): file/commit map, exact next actions, limitations, recovery, deployment evidence.
- [Runtime snapshot branch](https://github.com/puritysb/AgentDeck/tree/codex/dashboard-collaboration): `codex/dashboard-collaboration`.
- Own implementation commits: `a167b18e` (UI/firmware/transport), `41e0c486` (large-history read), `5477cb71` (verification).

### Do not merge the snapshot wholesale

The snapshot preserves pre-existing local Apple/Node changes in `ea6df63b` and `4979298f`; these are not this experiment's UI work. #284 and #285 subsequently merged reviewed versions of that work. Prepare integration in a new worktree based on current `master`, transplant only the experiment changes, and preserve the reviewed fixes. Keep the running snapshot intact for observation and rollback. No draft PR is opened merely to present this mixed-baseline snapshot as merge-ready.

## What is available to observe

- macOS title-bar switch: **기존 보기 / 협업 보기 · 실험**. Select a session in the existing roster/terrarium. The right rail shows its latest canonical task, typed child evidence and separately scoped live census; the network button retains access to the old topology rail. Existing approvals/timeline/terrarium remain.
- IPS10 only: equal-sized identity-ordered work cards within the supplied roster, parent state separate from optional child census, readable minimum height with scrolling. Individual child identities/dependency graphs are not implemented there. The inherited ten-session transport cap/rotation remains.
- Both transports add census only for identified IPS10 clients and remove the enhancement over the conservative 3,500-byte budget. Other board projections remain unchanged; missing census is unknown, not zero.

## Measured evidence (2026-09-06; local snapshot)

- Vitest: 259 files, 4,025 passed / 1 skipped.
[...]

PR #311, opened 2026-09-12
feat(experiment): an isolated attention-UI lab, and an opt-in read-only attention panel

puritysb:
Rebased onto `13a8f53d`. Parked in a local worktree since 2026-09-06; nothing in it had been pushed.

## What this is

Two commits, both **additive and opt-in** — no existing product default changes.

**1. `experiments/attention-lab/`** — a local browser A/B/C prototype over synthetic data, comparing how *terrarium presence*, *the moment a human is needed*, and *the context to come back to* relate. Runs on Node 22+ with no install, on `127.0.0.1:4317`, and never touches the root dev command or a real bridge.

**2. An opt-in menu-bar panel** — `Attention context (experimental)` under the gear, **off by default**, replacing only the existing right-hand Activity region. The terrarium, existing controls, daemon and protocol are untouched. Selection and existing row order are not reordered by new waits; a disconnect or a vanished selected session is stated as last-known context. It does not infer unobserved activity summaries, hook provenance, request expiry, or completion.

The first real-data use is the separate observation app in `experiments/attention-lab/native/`, which compiles the same SwiftUI/model sources directly and asks a verified existing loopback daemon for `dashboard-live/v1` `sessions.read` **only** — no agent command, no transcript storage, no notifications, no usage instrumentation, and the connection ends when the window closes. It deliberately did not replace the installed app or restart the daemon, to protect concurrent master work.

## Notes for review

- The branch predated the per-entry devlog migration and had edited the generated `DEVELOPMENT_LOG.md` directly. That content is now `docs/devlog/entries/2026-09-06-attention-ui-native-readonly-observation.md` and the aggregate is rebuilt from it.
- A third commit puts five `border-radius` values in the lab stylesheet onto the `{0,4,8,10,12,14,16,18,999}` design scale; without it the branch sat 5 over the lint baseline. The 6×6 dot moves `50%` → `999px`, which renders identically on a square element.
- `docs/appstore-feature-matrix.md` carries the row for the new panel, per the App Store rule that the matrix is updated before implementation.

## Verification

`pnpm build`, `pnpm typecheck`, `pnpm test` (281 files, **4413 passed / 1 skipped**), `pnpm generate-protocol` leaves **no drift**, `python3 design/verify-tokens-sync.py` all mirrors in sync, `pnpm docs:check`, `pnpm design-system:check`, `pnpm devlog:check` pass. `design/lint.sh --json` totals **89, equal to the baseline**.

From the original work: real session list received and manual selection confirmed; 19 pure-model checks passed; a final macOS Release build plus the App Store archive guard passed under local ad-hoc signing of the declared entitlements. **No usability conclusion has been drawn and no product default changed** — distribution signing, store submission and on-device popover QA are out of scope here.


puritysb:
Reviewed against the App Store invariants and the cross-surface rules. **The code is clean — my objection is to what it costs every shipped build for a feature nobody sees by default.** Holding rather than merging, and the fix is small.

## What I checked and found correct

- **No App Store violation.** No `Process()`, `/bin/sh`, `osascript` or `NSTask` anywhere in the Apple tree. `experiments/attention-lab/native/` is never referenced from `project.yml`, so `build.sh` and `ObservationApp.swift` do not enter any target. `validate-submission` passes.
- **`docs/appstore-feature-matrix.md` carries the row**, before implementation, and it correctly states the native host "is not bundled with or required by the App Store app".
- **Default-off wiring is right.** `@AppStorage(...) = false`, the toggle label is English, and the `if/else` swaps only `activitySummaryPanel`.
- **Font registration is lazy** — a `static let` reached from `AttentionContextPanel.init`, so a user who never opts in pays no launch cost.
- **Unknown agent is not laundered into another agent**: `row.agentType ?? "Unknown agent"`.

## Why I am not merging it

**1. 5.7 MB of fonts ship to every user for a default-off experiment.** `project.yml` adds `IBMPlexSansKR-Regular.ttf` (2.67 MB), `IBMPlexSansKR-Bold.ttf` (2.74 MB) and `JetBrainsMono-Regular.ttf` (0.26 MB) to the `AgentDeck_macOS` target. Lazy *registration* does not make them lazy *bytes* — they are in the bundle whether or not the toggle is ever flipped. These are the Korean-subset Plex faces, which is where the weight comes from.

**2. Twelve Korean-only strings enter an app whose shipped UI is 100% English.** `grep 'Text("[^"]*[가-힣]' apple/AgentDeck` returns **0 on master** and **12 on this branch** — `"실험 · 읽기 전용"`, `"필요할 때 맥락을 펼쳐보세요."`, and so on. There is no `.lproj`, no `String(localized:)`, no `NSLocalizedString` anywhere in the app, so these are hardcoded and would render Korean in every locale. That is also *why* the Korean font subset is needed — the two findings are the same decision.

**3. The default path is restructured for an off-by-default feature.** `overviewPanel` is now wrapped in `ScrollViewReader` and the `.background` moves inside it, for every user including those with the experiment off.

**4. This branch's own devlog says the experiment has not concluded** — "사용성 결론이나 제품 기본값 변경은 아직 하지 않았다". Nothing has yet earned the bundle weight.

## Suggested split

`experiments/attention-lab/` (commits 1 and 3) is pure addition, touches no target, and can land as-is. The Apple panel is worth holding until either the strings are English — which removes the Korean font requirement and both problems at once — or the panel uses system fonts while the experiment runs.

Happy to do the English-strings version if you want it landed now.

puritysb:
Closing without merging. I went back to why this was raised, and the branch answers it itself.

**Its own gate was never reached.** The lab README states the next decision explicitly — *"이 실험을 곧바로 병합하는 것이 아니라 실제 사용자 관찰로 A/B/C 가설을 좁힌 뒤 제품 내부 opt-in 패널로 옮길지 정한다"* — and adds *"수치와 선호도는 아직 수집하지 않았습니다"*. The native README asks for notes from the first 3–5 real interventions. Six days on there is no such record in the devlog, the issues, or here. The instrument was built; the measurement never ran.

**The product answered the question meanwhile.** `AttentionTheaterView` already owns "the moment a human is needed" — and it renders *answerable* options, which this panel deliberately does not. And `activitySummaryPanel`, the region this would replace, has since become an APME-backed `MenuBarActivityGlance` with a Full report link, a daemon-offline empty state and accessibility labels. Swapping it for a read-only roster would be a downgrade, not an experiment. The branch is 64 commits behind, so it was designed against a master that no longer exists — the same shape as `codex/dashboard-collaboration`, which I retired today for the same reason.

The sibling experiment #287 (collaboration lens / IPS10 work cards) completed the observe→hand off→land cycle and closed. Between them, the "work list" direction of A/B/C was effectively settled by shipping.

**The thinking is preserved**, not thrown away: `docs/devlog/entries/2026-09-12-attention-lab-experiment-retired.md` (`26f53e55`) records the five axes worth keeping stable across surfaces (대상 · 관측한 상태 · 요청 · 출처 · 응답 가능 여부), the rule that an unknown request kind is never assumed answerable, what merging would have cost, and the lesson — **give an experiment a measurement deadline, not just an instrument.**

Nothing is lost by closing: no production code here ever ran for a user, and the branch is recoverable from this PR if the A/B/C review is ever actually run.

PR #329, opened 2026-09-14
fix(hooks): invoke the Windows hook as a script file, not an inline -Command

### LINKED ISSUE #331: Windows: every Claude Code hook fails — Git Bash strips $variables from the inline PowerShell -Command
## Summary

On Windows, every AgentDeck Claude Code hook fails with a PowerShell parse error, on every event, on every turn. Hook-based observation never reaches the daemon, and the user sees a wall of errors in the Claude Code UI.

## What I saw

```
Stop hook error: Failed with non-blocking status code: At line:1 char:57
+ ='Stop'; [int]=0; [int]=0; if(!([int]::TryParse([string]:AGENTDECK_PO ...
+                                                         ~
Missing ')' in method call.
At line:1 char:57
+ ... int]=0; [int]=0; if(!([int]::TryParse([string]:AGENTDECK_PORT,[ref])) ...
+                                                   ~~~~~~~~~~~~~~~
Unexpected token ':AGENTDECK_PORT' in expression or statement.
...
At line:1 char:123
+ ... TDECK_PORT,[ref])) -or  -lt 1 -or  -gt 65535){=0}; =; if(-not ){=Join ...
+                                                                  ~
Missing expression after unary operator '-not'.
Not all parse errors were reported.  Correct the reported errors and try again.
    + FullyQualifiedErrorId : MissingEndParenthesisInMethodCall
```

## Environment

- Windows 11 Pro 10.0.26200
- Node v22.23.2, `@agentdeck/bridge` 1.3.3 (global, `C:\nvm4w\nodejs`)
- Hooks installed to `~/.claude/settings.json` by the current installer

## Diagnosis

Compare what is stored in `~/.claude/settings.json`:

```
powershell -NoProfile -ExecutionPolicy Bypass -Command "$ev='Stop'; [int]$port=0; [int]$candidate=0; if(!([int]::TryParse([string]$env:AGENTDECK_PORT,[ref]$candidate)) ..."
```

with what PowerShell reports receiving:

```
='Stop'; [int]=0; [int]=0; if(!([int]::TryParse([string]:AGENTDECK_PORT,[ref])) ...
```

Every `$name` is gone, and `$env:AGENTDECK_PORT` became the literal `:AGENTDECK_PORT`.

That is POSIX parameter expansion, not a PowerShell problem. **Claude Code spawns hook commands through Git Bash (`sh -c`) on Windows**, so `$ev`, `$port`, `$candidate` and `$env` are expanded to empty strings before `powershell.exe` ever parses the line; `:AGENTDECK_PORT` survives because `:` terminates a shell variable name. Double-quoting inside the settings JSON does not help — the shell sees the whole command string as one argument.

The assumption is stated in the source. [`hooks/src/install.ts`](https://github.com/puritysb/AgentDeck/blob/master/hooks/src/install.ts) on the builder's doc comment:

> Claude Code v2.1+ executes hook commands through `cmd.exe` on Windows […] Single quotes are used inside the PowerShell script so the entire `-Command` argument can stay double-quoted **under cmd.exe**.

It is not cmd.exe.

## Reproduce

On Windows, with hooks installed by `node hooks/dist/install.js` or `npx @agentdeck/setup`:

```sh
# what Claude Code effectively does
CMD=$(node -e "console.log(require('./hooks/dist/install.js').buildHookCommandWin('Stop'))")
echo '{}' | sh -c "$CMD"
```

→ the parse errors above. The same command run directly from PowerShell or cmd.exe works, which is why this was not caught.

## Impact

- Hook-based state never reaches the daemon on Windows (`SessionStart`, `UserPromptSubmit`, `PreToolUse`, `Stop`, …).
- The errors are surfaced to the user on every turn.
- Affects the `@agentdeck/hooks` installer, `@agentdeck/setup`, and the Kiro hook file, which all route through the same builder.

## Also affected (found while fixing)

- **Ownership detection.** `applyHooks` / `removeHooks` / the legacy sweeps identify AgentDeck hooks by `AGENTDECK_PORT` or `localhost:9120`. Any hook shape carrying neither is invisible to them, so a reinstall appends instead of replacing and each event fires once per install.
- **No migration catches this.** Migration 9's `[int]::TryParse` marker matches the broken form as well as a fixed one, so an existing install never self-heals.

## Fix

PR #329 — ships the hook body as `~/.agentdeck/agentdeck-hook.ps1` and invokes it with `-File`, so no `$` reaches the command line. Port discovery, strict `1..65535` parsing, UTF-8 stdin/POST, the 2s timeout and `exit 0` are unchanged. Adds `isAgentDeckHookCommand()` in one place and migration 12 to repair affected installs.

dudziakl:
Fixes #331

## The bug

Every AgentDeck hook on Windows fails, on every event, on every turn:

```
Stop hook error: At line:1 char:57
+ ='Stop'; [int]=0; [int]=0; if(!([int]::TryParse([string]:AGENTDECK_PO ...
+                                                         ~
Missing ')' in method call.
```

Look at what `powershell.exe` actually received: every `$name` is gone, and `$env:AGENTDECK_PORT` arrived as the literal `:AGENTDECK_PORT`.

That is POSIX parameter expansion, not a PowerShell problem. **Claude Code spawns hook commands through Git Bash (`sh -c`) on Windows**, so `$ev`, `$port`, `$candidate` and `$env` expand to empty strings before PowerShell parses the line — `:AGENTDECK_PORT` survives because `:` terminates a shell variable name. Double-quoting inside the settings JSON does not help; the shell sees the whole command string.

`buildHookCommandWin`'s doc comment named the wrong host:

> Single quotes are used inside the PowerShell script so the entire `-Command` argument can stay double-quoted **under cmd.exe**.

Net effect: hook-based observation is dead on Windows, and the user sees a wall of PowerShell parse errors on every turn.

## The fix

Ship the hook body as a script file and invoke it by path, so no `$` ever reaches the command line:

```
powershell -NoProfile -ExecutionPolicy Bypass -File "<home>\.agentdeck\agentdeck-hook.ps1" -HookEvent <Event>
```

- `WINDOWS_HOOK_SCRIPT` holds the body; `ensureWindowsHookScript()` writes it idempotently to `~/.agentdeck/` and repairs a tampered copy.
- Behaviour is unchanged: same port discovery (`$env:AGENTDECK_PORT` → `daemon.json` + `/health` probe → `9120`), same strict `1..65535` parse, same UTF-8 stdin read and UTF-8 byte POST, same 2s timeout, same `exit 0`.
- `ValidatePattern` accepts `_` so the prefixed Kiro daemon events (`kiro_tool_start`, …) still pass.
- `setup/src/setup.ts` carries the byte-identical copy it has to — that package bootstraps its own deps and cannot import `@agentdeck/hooks`. Swift `HookInstaller` has no Windows path to mirror.

## Two consequences worth reviewing

**1. Hook ownership detection.** The `-File` command contains neither `AGENTDECK_PORT` nor `localhost:9120` — the two markers every writer used to decide "this entry is ours". Left alone, each reinstall would keep the old entry and append a new one, so every event fires once per install. That test is now `isAgentDeckHookCommand()`, defined once and used by `applyHooks`, `removeHooks` and both legacy sweeps, matching all three shapes.

**2. Migration 12.** Migration 9's `[int]::TryParse` marker matches the broken *and* the fixed form, so no existing migration catches this. Migration 12 rewrites any Windows settings file that does not mention `agentdeck-hook.ps1`. `uninstallHooks` now removes the script, the way the OpenCode plugin and Kiro hook file are removed.

`applyHooks`/`buildHookEntry` take an optional `home` so the emitted absolute path matches the home actually being installed into (defaults to `homedir()`, so no call site changes).

## Verification

- Generated command run through `sh -c` with a payload on stdin — the exact path Claude Code uses: **exit 0, no parse errors**.
- `pnpm build`, `pnpm typecheck` clean.
- `pnpm test`: 4502 passed. The 16 failures on this Windows host are identical to those on unmodified `master` here (missing `python3`, macOS-only chmod/lipo paths, systemd, and `generate-protocol.sh` resolving to a WSL `bash` without `node`); no `hooks/` or `setup/` test among them.
- `hooks/src/__tests__/install.test.ts`: 49 passed, 3 skipped (POSIX-only). New coverage for the no-`$` invariant across every event, the script's preserved behaviour, the Kiro event pattern, idempotent/repairing writes, `isAgentDeckHookCommand`, migration 12, no duplication across repeated installs, and script removal on uninstall.
- `pnpm docs:check`, `pnpm design-system:check`, `pnpm devlog:build` + `:check` pass.

`pnpm generate-protocol` could not run on this host (`scripts/generate-protocol.sh` resolves to a WSL `bash` that has no `node` — same failure on unmodified `master`). The change touches no protocol types, so no drift is possible.

Devlog entry added at `docs/devlog/entries/2026-09-14-windows-hook-script-file.md`; `docs/windows.md` and the `CLAUDE.md` Windows line updated to describe the script-file form.

https://claude.ai/code/session_01JGySVbSFf4RtsUCpuidBkG

## Maintainer follow-up

The original Windows diagnosis and commit are preserved. Follow-up `177e4cb4` repairs standalone Kiro script provisioning (including the already-current path), forwards the supplied home, and converges Windows settings by value rather than searching them for validation code that moved into the script. Missing/corrupt scripts are repaired independently of settings rewrites. The bootstrap script has a byte-identity gate.

Local macOS validation: full build/typecheck; 291 test files, 4,504 tests passed and 1 existing skip before adding the Windows runtime integration case; the final installer suite passes 55 tests with its real-Windows integration case skipped on macOS. Protocol generation has no drift; docs/catalog/devlog/tokens pass; clean-source design lint is 89, identical to the base.

The Windows CI matrix now runs installer/observer/state-machine tests on Node 22, 24 and 26, including an actual Git Bash -> Windows PowerShell -> local HTTP round trip with Korean/UTF-8 payload and a Unicode/spaced script path. All final CI checks passed before merge. No package release or installed-daemon upgrade is implied by this PR.

### Combined candidate after #304 and #330

`7ae6627f` incorporates the two accepted PRs, preserves the direct-checkout instructions, and regenerates the devlog from both source entries. The combined local tree passes build/typecheck, all 291 test files (4,515 tests passed, 2 skipped), docs/catalog/devlog checks, and protocol/state-transition generation with no drift.

The previous candidate's actual Windows Git Bash/PowerShell HTTP integration passed on Node 24 and 26, and on Node 22 after one unchanged rerun. The first Node 22 attempt hit the test's 10-second process budget; the rerun's installer suite completed in 3.45 seconds. The final combined candidate passed every CI check before merge, including coverage and Windows Node 22/24/26 (174 targeted tests passed per Windows version).


hooks/src/install.ts
```diff
@@ -115,37 +115,101 @@ export function buildHookCommand(eventName: string): string {
...
+}
+
+/**
+ * Windows variant: invoke `agentdeck-hook.ps1` by path. The command line
+ * carries no `$`, so the POSIX shell Claude Code spawns hooks through has
+ * nothing to expand. See WINDOWS_HOOK_SCRIPT for why that matters.
+ */
+export function buildHookCommandWin(eventName: string, home: string = homedir()): string {
+  const script = windowsHookScriptPath(home);
+  return `powershell -NoProfile -ExecutionPolicy Bypass -File "${script}" -HookEvent ${eventName}`;
```
puritysb, line 212: The new file dependency must also be provisioned by installKiroHooksIfNeeded(): buildKiroHookFile() uses this builder, but the standalone Kiro installer never calls ensureWindowsHookScript(). In a fresh home it reports installed while the referenced script does not exist. Please provision/repair the script in the Kiro installer and pass its home through buildKiroHookFile() to this builder. A regression should cover both first install and the already-current hook-file path.

puritysb (changes requested):
Thank you for tracing this through Git Bash and testing on the affected Windows host. Moving the PowerShell body into a file addresses the reported failure, and the ownership detection update is necessary.

I reproduced one installation gap below. I also found that migration 9 still searches settings.json for [int]::TryParse, which now lives only in the .ps1 file: a current Windows install therefore rewrites settings on every migration call. Script repair should be independent of settings migration.

The existing installer suite passes here (49 passed, 3 platform skips); a standalone-Kiro Windows-branch regression fails. This was tested on macOS with the Windows branch selected, not claimed as Windows runtime verification.

Maintainer edits are enabled. I will add small follow-up commits on this PR, preserving your original commit, with regressions and full validation before merging. Thank you for contributing the diagnosis and fix.

puritysb (approved):
All findings are addressed in the final combined candidate 7ae6627f: Kiro provisions and repairs its script independently, the configured home reaches every command, current Windows settings are not rewritten on session start, and the bootstrap script is byte-checked.

Validation is now complete: local build/typecheck and 4,515 tests pass; protocol/state-transition generators have no drift; docs/catalog/devlog/tokens pass. Every final CI check is green, including coverage and Windows Node 22, 24 and 26. On each Windows version the targeted suite passes 174 tests (3 POSIX-only skips), including the actual Git Bash -> Windows PowerShell -> local HTTP round trip with a Korean payload and a Unicode/spaced script path.

The original diagnosis, runtime evidence, and contributor commit are preserved alongside the maintainer follow-ups. Thank you for finding and fixing the Windows hook failure. Ready to merge; npm publication and deployed-daemon upgrades are separate from this source merge.

PR #330, opened 2026-09-14
fix(bridge): stop blanking the observed roster and latching the hub at disconnected

### LINKED ISSUE #332: Windows: observed-session roster flashes then empties, and the hub latches at disconnected
## Summary

On a Windows 11 host the dashboard showed the observed sessions for a few seconds, then went empty, repeatedly — and `GET /status` reported `disconnected` the entire time work was running. Two independent defects.

## What I saw

> from time to time it works and shows 4 sessions […] but they are displayed for a few seconds and disappear.

`sessions_list` broadcasts alternated between all sessions and none, roughly on the 5s scan interval. `curl 127.0.0.1:9120/status` returned `"state": "disconnected"` while four Claude Code sessions were actively running tools. Earlier in the same session it had been latched at `idle` instead.

## Environment

- Windows 11 Pro 10.0.26200
- Node v22.23.2, `@agentdeck/bridge` 1.3.3 (global), daemon on 9120
- 4 live `claude.exe` sessions

---

## Defect 1 — a failed process scan is treated as "every session ended"

Measured on that machine, four consecutive runs:

```
Get-CimInstance Win32_Process | Select-Object ProcessId,ParentProcessId,WorkingSetSize,CommandLine  →  ~10.4s
powershell.exe -NoProfile -NonInteractive -Command "1"                                              →  0.55s
```

So it is the WMI provider, not PowerShell startup, and narrowing the WQL projection does not help. Against `collectProcessInfoWin32`'s `timeout: 10_000`, the query **timed out on every single scan**.

The timeout alone is not what emptied the roster. `collectProcessInfo()` returns `[]` on a timeout, a spawn failure or unparseable output rather than throwing — so the `.catch(() => { this.cached = []; })` in `PassiveSessionObserver.collect()` never fired, and the empty list flowed straight through `scan()`, replacing the roster with nothing. Every client saw every observed session end at once, then reappear if a scan happened to beat the clock.

Secondary: `SCAN_INTERVAL_MS` is 5s and `lastScanAt` is stamped when a scan *starts*, so an 11s scan means the next one begins the moment the last finishes — the machine runs the WMI query back-to-back forever.

## Defect 2 — the hub has no edge back to `PROCESSING`

The daemon hub constructs its state machine with `toolActivityRecovery: false` (`bridge-core.ts`) so one observed session's tools cannot dismiss another session's `AWAITING_*` prompt. But `recoverProcessingFromToolActivity()` returns on that flag before *every* state, not only those.

Combined with the wildcard `{ from: '*', to: DISCONNECTED, trigger: 'session_end' }` row — which fires whenever **any** observed session ends — the hub's only remaining way out was `session_start`, a hook that never fires again for a session already underway. So one session ending stranded the whole hub, and nothing could reopen it.

Related: there is no `PROCESSING → PROCESSING` row, so `transition()` drops tool activity that arrives while already processing. For a hook-only (observed) session there are no PTY bytes, so `onPtyActivity()` never runs either — the stuck timer is never re-armed and a long turn decays to `IDLE` after `STUCK_TIMEOUT_MS` however busy it is.

---

## Reproduce

Defect 1 does not need slow WMI — any scan failure reproduces it:

```js
// PassiveSessionObserver.scan() with collectProcessInfo() returning []
// → this.cached = [] → sessions_list broadcasts an empty roster
```

Defect 2:

```js
const sm = new StateMachine(tracker, { toolActivityRecovery: false });
sm.handleHookEvent('SessionStart', {});
sm.handleHookEvent('SessionEnd', {});            // any one session ending
sm.handleHookEvent('PreToolUse', { tool_name: 'Bash', tool_input: {} });
sm.getState();                                    // 'disconnected' — and stays there
```

## Impact

- Windows hosts with a slow `Win32_Process` provider get a roster that flashes and vanishes, and the daemon runs WMI continuously.
- Any host where a scan fails transiently publishes a momentarily empty roster.
- The hub's global state latches at `disconnected` (or `idle`) once any session ends, so `/status` and anything driven from it are wrong for the rest of the daemon's life.

## Fix

PR #330 — retains the roster when the process table comes back empty or the scan rejects, widens the Windows budget to 30s, and makes `SCAN_INTERVAL_MS` a floor with an adaptive back-off (fast hosts unchanged at 5s). Narrows the hub guard to `AWAITING_*`/`PROCESSING`, adds `DISCONNECTED → PROCESSING` on `tool_activity` to the shared transition SSOT, and re-arms the stuck-timer backstop on tool activity during `PROCESSING`.

## Not addressed

The slow-WMI root cause itself. Worth chasing separately on an affected machine: an elevated `winmgmt /verifyrepository`, and whether an AV/EDR product is hooking the `Win32_Process` provider.

dudziakl:
Fixes #332

Two independent defects found on the same Windows 11 host: the dashboard showed four observed sessions for a few seconds, then emptied, on repeat, while `GET /status` reported `disconnected` the whole time work was running.

Independent of #329 (Windows hooks) — different files, no overlap.

## 1. The observed roster blanked on every scan

`Get-CimInstance Win32_Process` measured **~10.4s** on that machine across four runs. Bare `powershell.exe -NoProfile` startup there is 0.55s and narrowing the WQL projection does not help, so it is the WMI provider itself. Against `collectProcessInfoWin32`'s 10s budget, that timed out on **every single scan**.

The timeout is not what emptied the roster, though. `collectProcessInfo()` reports a timeout, a spawn failure or unparseable output as `[]` rather than throwing — so `.catch(() => { this.cached = []; })` never fired, and the empty list flowed straight through `scan()` as "every observed session ended".

Three changes, and a fourth for the cost:

- **Empty process table is retained, not believed.** No running machine has zero processes, so `[]` is the third answer — "I could not look" — and the previous roster stands. This is the actual blanking path.
- **A rejected scan keeps the last roster too.** A failed scan is not evidence that every session ended.
- **Windows budget 10s → 30s**, with the measurement in the comment.
- **`SCAN_INTERVAL_MS` stays 5s but becomes a floor.** An 11s scan on a 5s interval starts the next one the instant the last finishes, pinning the machine at 100% of whatever the process table costs. `nextScanIntervalMs()` backs off to the last scan's own duration, capped at 60s — so that host sits at a ~50% duty cycle and fast hosts (a few hundred ms on macOS/Linux) are unchanged at 5s.

`PassiveSessionObserver` takes an optional injected process-table reader, mirroring `CodexRolloutCache`'s existing dependency seam, so the failure paths are testable without a real process table.

## 2. The hub had no edge back to PROCESSING

Separate defect, same session. The daemon hub runs with `toolActivityRecovery: false` so one observed session's tools cannot dismiss another session's `AWAITING_*` prompt — but the guard returned before *every* state, not just those. Combined with the wildcard `session_end → DISCONNECTED` row, which fires whenever **any** session ends, the hub had only `session_start` as a way out — a hook that never fires again for a session already underway. It latched.

- The guard now covers `AWAITING_*`/`PROCESSING` only. On a machine multiplexing every session, "some session started a tool" makes `PROCESSING` the truthful aggregate, so `IDLE`/`DISCONNECTED` can recover. A straggler tool **END** still cannot reopen a finished session.
- `shared/src/states.ts` gains `DISCONNECTED → PROCESSING` on `tool_activity`, with its rationale as the `note`; regenerated into `StateTransitions.generated.swift` via `pnpm generate-state-transitions` (no drift).
- Tool activity during `PROCESSING` re-arms the stuck-timer backstop the way `onPtyActivity()` does. There is no `PROCESSING → PROCESSING` row, so `transition()` dropped these and a hook-only session — no PTY bytes, so `onPtyActivity()` never runs — decayed to `IDLE` after `STUCK_TIMEOUT_MS` however busy it was. Silence still recovers; the backstop is re-armed, not disarmed.

### One contract change to review

The existing test *"the daemon hub machine never dismisses prompts or reopens IDLE on tool activity"* asserted the old behaviour. I split it: the hub still never dismisses a held prompt (unchanged, still tested), but it now does reopen from `IDLE`/`DISCONNECTED` on a tool **start**. If you consider the hub's `idle` more truthful than `processing` while other sessions work, this is the line to push back on.

## Verification

The runtime evidence is from the equivalent patch applied to the installed `@agentdeck/bridge@1.3.3`, not from this branch: **10 consecutive `sessions_list` broadcasts over 75s, every one carrying all 4 sessions**, and `/status` moving from `disconnected` to `processing`. Before the patch the roster emptied on essentially every scan.

On this branch:

- `pnpm build`, `pnpm typecheck` clean; `pnpm generate-state-transitions` reports no drift.
- `pnpm test`: 4501 passed. The 16 failures on this Windows host are identical to those on unmodified `master` here (missing `python3`, macOS-only chmod/lipo paths, systemd, and `generate-protocol.sh` resolving to a WSL `bash` without `node`); `passive-observer` and `state-machine` pass fully.
- 6 new tests. Mutation-checked: 2 of the 3 observer tests fail without the source fix, and the third fails if the `cached = []` line is reinstated.
- `pnpm docs:check`, `pnpm design-system:check`, `pnpm devlog:build` + `:check` pass. `bash design/lint.sh` reports 91 violations, all pre-existing in `docs/*.html` and none in files this branch touches.

## Not fixed here

The slow-WMI root cause is accommodated, not solved. Worth chasing separately on an affected machine: an elevated `winmgmt /verifyrepository`, and whether an AV/EDR product is hooking the `Win32_Process` provider.

Devlog entry at `docs/devlog/entries/2026-09-14-windows-observer-roster-and-hub-recovery.md`.

https://claude.ai/code/session_01JGySVbSFf4RtsUCpuidBkG

## Maintainer follow-up

The original contribution and Windows measurements are preserved. Follow-up `edf212b4` fixes the two issues found in review: slow scans cool down from completion, while fast scans preserve the 5-second start-to-start cadence; PROCESSING liveness refresh executes before the hub's prompt-dismissal guard. Tests cover actual scheduling for fast/slow/rejected/capped scans, overlapping collect calls, and the stuck timer with both session and hub configurations.

Local macOS validation: build/typecheck; all 291 test files, 4,505 tests passed and 1 skipped; observer/state-machine suites pass 121 tests. Protocol and state-transition generators have no drift. Docs/catalog/devlog/tokens pass, and clean-source design lint is 89, identical to base. The author's runtime evidence above remains evidence from their Windows host, not a claim that this exact follow-up was deployed there.

All CI checks passed before merge, including native Windows 22/24/26, macOS tests, and iOS build. No package release or installed-daemon upgrade is implied by this PR.


bridge/src/passive-observer.ts
```diff
@@ -252,18 +278,32 @@ export class PassiveSessionObserver {
    */
   collect(managedSessions: EnrichedSession[]): ObservedSession[] {
     const now = Date.now();
-    if (now - this.lastScanAt >= SCAN_INTERVAL_MS && !this.scanInFlight) {
+    if (now - this.lastScanAt >= nextScanIntervalMs(this.lastScanMs) && !this.scanInFlight) {
```
puritysb: lastScanAt is the scan START time, so when an 11-second scan completes, now - lastScanAt already equals nextScanIntervalMs(11000). The next collect() immediately starts another scan, retaining near-continuous WMI work rather than the documented cooldown. I reproduced two reader calls by advancing one 11-second scan and calling collect() immediately after completion. Schedule the cooldown from completion (while preserving the fast-host cadence) and test collect(), not only the interval helper.

bridge/src/state-machine.ts
```diff
@@ -542,30 +542,47 @@ export class StateMachine extends EventEmitter {
...
+   *  - IDLE / DISCONNECTED: only a tool START recovers a dropped
+   *    user_prompt_submit. A straggler tool-END curl landing after Stop must
+   *    not reopen a finished session (nothing would ever re-close it).
    *
    *  A dismissal skipped by any guard is retried by the next hook — a
    *  keyboard-answered prompt always produces later tool activity or a stop. */
   private recoverProcessingFromToolActivity(kind: 'start' | 'end'): void {
-    if (!this.toolActivityRecovery) return;
+    if (!this.toolActivityRecovery
+      && this.state !== State.IDLE && this.state !== State.DISCONNECTED) return;
```
puritysb: For the daemon hub (toolActivityRecovery: false), this guard returns while PROCESSING before the new armStuckTimer branch can run. A hub receiving PreToolUse every 10 seconds still becomes IDLE at STUCK_TIMEOUT_MS. Refresh liveness before the hub prompt-dismissal guard, retaining the AWAITING_* protections, and exercise the hub configuration in the regression.

puritysb (changes requested):
Thank you for the measured WMI diagnosis and for distinguishing unavailable process data from an empty roster. The cache-retention change addresses the observed blanking path.

I reproduced two gaps below using the actual observer and hub state machine with fake timers. The existing observer/state-machine suites pass (116 tests), while both added behavioral probes fail. The current backoff unit test exercises the arithmetic helper but not scheduling; the stuck-timer test uses the default session machine rather than the hub configuration.

Maintainer edits are enabled. I will add focused follow-up commits to this PR, preserving your original commit, then run the full checks before merging. Your Windows runtime measurements remain valuable; my local verification is on macOS.

puritysb (approved):
Both review findings are fixed in edf212b4 with behavioral regression tests: the real scan scheduler now waits after slow attempts, and the hub refreshes its PROCESSING timer without dismissing held prompts. Local build/typecheck and 4,505 tests pass; protocol/state-transition generators are synchronized. All current CI checks are green, including Windows native runtime on Node 22/24/26, macOS tests and iOS build. Thank you for the Windows measurements and the original fix; both original and maintainer commits are retained.

PR #334, opened 2026-09-14
release(npm): recover blocked publication with 1.3.5

puritysb:
Deliver the reviewed Windows fixes as npm 1.3.5 after the registry wedged shared 1.3.4: initial publish succeeded, public lookup stayed 404, retry returned “previously staged version”, and authenticated staging lists are empty (same symptom as npm/cli#9889). Hooks/bridge/setup 1.3.4 are already immutable; preserve them and the existing tag, and publish a complete fresh four-package set.

Only versions, the version fixture, changelog and delivery evidence change. Runtime fixes are unchanged from #333. Build/typecheck and 4,515 tests passed (2 platform skips), protocol generation has no drift, docs/catalog/devlog/release-version checks passed. The existing installed three-mode results are documented; the new exact-commit candidate is being installed and verified before its tag.


puritysb:
The registry has now exposed shared@1.3.4, and all four exact versions/latest tags have been independently verified. Withdraw this unused replacement version; finish the original immutable npm-v1.3.4 release. No 1.3.5 package or tag was published.

Issue #46, opened 2026-07-06
fix(windows): make Claude Code hooks preserve UTF-8 payloads


puritysb:
Follow-up from PR #43. The Codex Windows hook path now reads stdin as UTF-8 and posts UTF-8 bytes, but the Claude Code Windows one-liner in hooks/src/install.ts still uses [Console]::In.ReadToEnd() and posts a string body through Invoke-RestMethod. On Windows PowerShell 5.1 this can decode piped stdin through the OEM codepage and encode request bodies as ISO-8859-1, mangling non-ASCII hook payloads.\n\nAcceptance criteria:\n- Update buildHookCommandWin / Claude Windows hook generation to read stdin through a UTF-8 StreamReader.\n- Post UTF-8 bytes with application/json; charset=utf-8.\n- Add Windows-generation tests with non-ASCII payload expectations, mirroring the Codex coverage added in #43.\n- Preserve existing reinstall migration behavior so AgentDeck-managed Claude hook entries self-update.

PR #336, opened 2026-09-14
fix(windows): run the autostart daemon with no console window

### LINKED ISSUE #339: Windows: daemon autostart leaves a permanent console window (Task Scheduler action is the daemon itself)
## Symptom

On Windows 11, `agentdeck daemon install` leaves a terminal window on the desktop — with a taskbar button — for as long as the daemon runs. The window shows the daemon's whole startup log and closing it kills the daemon. It comes back at every logon.

```
C:\WINDOWS\system32\cmd
…
[daemon] Daemon running (build 9f8897016751)
```

Reproduced on Windows 11 Pro 26200, Node 22.23.2, npm-installed CLI (`npx @agentdeck/setup`), Windows Terminal as the default terminal application.

## Root cause

The Scheduled Task `AgentDeckDaemon`'s action was `daemon start --foreground`, so **the task's process *is* the daemon**. Task Scheduler starts an `InteractiveToken` action with a console attached, and `node.exe` is a console-subsystem binary, so the logon start hands the daemon to the default terminal application.

There is no window to hide: `EnumWindows` finds the window owned by `WindowsTerminal` (titled with the action's command line) while the daemon's own `node.exe` reports `MainWindowHandle = 0`, and that same Terminal process also hosts the user's unrelated PowerShell windows — so `ShowWindow` would hide the user's console too.

Three plausible fixes were measured and do **not** work:

| Candidate | Result |
|---|---|
| `<Hidden>true</Hidden>` in the task XML | Probe task still produced a visible `WindowsTerminal` window. The setting hides the task in the Task Scheduler UI, nothing else. |
| `conhost.exe --headless node …` | Emitted terminal escape sequences into the calling console and never ran the target script. |
| `S4U` principal ("run whether the user is logged on or not") | `schtasks /Create` → `ERROR: Access is denied.` without elevation, which `daemon install` deliberately does not have. It would also return the daemon to a non-interactive session — the reason a real Windows Service was rejected for this job in the first place. |

## Solution (implemented in #336)

**Make the task's action a launcher instead of the daemon.** The action becomes `daemon autostart`, a hidden CLI command that spawns `daemon start --foreground` with `detached` + `windowsHide` (`DETACHED_PROCESS | CREATE_NO_WINDOW`) and exits within a few hundred ms — before the terminal handoff paints anything. A probe task doing exactly this showed no visible window at 0.7 s or at 5 s after `schtasks /Run`, with the detached daemon still serving.

That change has three consequences that have to be handled together, and each one is a bug on its own if it is missed:

1. **The task's status stops describing the daemon.** `Status: Ready` becomes the steady state of a *healthy* machine. Read as "not running", it makes `classifySupervision` return `unsupervised`, whose remedy is stopping the daemon — so the window fix alone would make every install and restart stop a healthy daemon. Fix: the launcher passes `AGENTDECK_SUPERVISOR=schtasks` and **the daemon** stamps `startedBy` into `daemon.json` past its successful bind; `composeSchtasksRunning` reads that stamp when the status says `Ready`. The stamp must come from the daemon, not the launcher — a launcher-recorded pid reported a daemon that was alive only because it was *conceding* to an incumbent as supervised.

2. **`daemon install` must end a running task instance before `/Run`.** With `MultipleInstancesPolicy=IgnoreNew`, `schtasks /Run` against an instance of the *previous* action is a silent rc-0 no-op — so on the upgrade the newly registered launcher never ran and the console-window daemon kept serving behind an install that reported success.

3. **Every child the daemon spawns now needs `windowsHide`.** A console-subsystem child of a console-*less* parent gets a brand new console, and a new console comes with a window. This surfaced as an empty `C:\Windows\system32\taskkill.exe` window appearing at random (the Codex rate-limit probe's five-minute cleanup kill: `codex.cmd` runs under `shell`, so the child is `cmd.exe` and the server is its grandchild, hence `taskkill /T`). Measured from a console-less detached parent, 12 spawns per arm: **138 window sightings without `windowsHide`, 0 with it**. The `adb reverse` calls in the daemon's poll loop had the same gap. A vitest gate (`bridge/src/__tests__/windows-child-window.test.ts`) now fails any `child_process` call that neither hides its window, nor runs a binary absent on Windows, nor carries an inline `windows-hide-exempt:` reason.

The opposite fix does not exist: a child spawned with `windowsHide` but *not* `detached` inherits the launcher's console (CREATE_NO_WINDOW is ignored when no new console is created) and dies with it. So the daemon stays console-less and the spawns are what change.

Two losses, stated rather than hidden: `RestartOnFailure` now covers only a launcher that cannot spawn (a crashed daemon returns at the next logon or by hand), and `schtasks /End` no longer stops the daemon — `daemon stop` was already ending the task *and* POSTing `/shutdown`, and the second one is what did the work.

## Upgrade note for anyone testing this

The patched build must be the one the **installed** CLI carries. A reboot regressed to the visible window because `agentdeck daemon install` had been run from the *published* 1.3.4 CLI, which has no `daemon autostart` command and can only write the action whose process is the daemon. Any `daemon install` or `npx @agentdeck/setup` from an unpatched build silently restores the window at the next logon.

## Verified

Daemon starts at logon with **no console window and no taskbar button**, `daemon.json` carries `"startedBy": "schtasks"`, `/health` answers `mode: daemon`, task status `Ready`, `daemon status` / `stop` / `start` / `restart` all behave, and no stray child windows since the `windowsHide` fixes. Confirmed across a full reboot.

Fix: #336

dudziakl:
Closes #339.

## The problem

On Windows, `agentdeck daemon install` leaves a terminal window — and a taskbar button — on the desktop for as long as the daemon runs. The Scheduled Task's action was `daemon start --foreground`, so the task process *is* the daemon, and Task Scheduler attaches a console to an interactive-token action with no way to suppress it.

Reported and reproduced on Windows 11 26200 with Windows Terminal as the default terminal app. `EnumWindows` shows the window owned by `WindowsTerminal`, titled with the action's command line, while the daemon's own `node.exe` reports `MainWindowHandle = 0` — and that same Terminal process hosts the user's unrelated consoles, so there is no window belonging to the daemon that could be hidden or closed.

## Alternatives measured and rejected

| Candidate | Result |
|---|---|
| `<Hidden>true</Hidden>` in the task XML | A probe task carrying it still produced a visible `WindowsTerminal` window. The setting hides the task in the Task Scheduler UI, nothing more. |
| `conhost.exe --headless node …` | Emitted escape sequences into the calling console and never ran the target script. |
| `S4U` principal ("run whether the user is logged on or not") | `schtasks /Create` → `ERROR: Access is denied.` without elevation, which `daemon install` deliberately does not have. It would also return the daemon to a non-interactive session — the reason a real Windows Service was rejected for this job. |

## The fix

The action becomes `daemon autostart`, a hidden CLI command that spawns `daemon start --foreground` with `detached` + `windowsHide` (DETACHED_PROCESS | CREATE_NO_WINDOW) and exits within a few hundred ms, before the terminal handoff paints anything. A probe task doing exactly this showed **no visible window at 0.7 s or at 5 s** after `schtasks /Run`, with the detached daemon still serving.

Running the install for real surfaced two consequences, both fixed here rather than left for the next reader:

1. **The task's status stops describing the daemon.** `Status: Ready` becomes the steady state of a *healthy* machine, and reading it as "not running" makes `classifySupervision` return `unsupervised` — whose remedy is stopping the daemon — on every install and restart. The daemon answers instead: the launcher passes `AGENTDECK_SUPERVISOR=schtasks`, the daemon stamps `startedBy` into `daemon.json` **past its successful bind**, and `composeSchtasksRunning` reads that stamp when the status says `Ready`. The stamp has to come from the daemon: a first version had the launcher record the pid it spawned, and on the first real install that reported a daemon which was alive only because it was *conceding* to an incumbent as supervised — the install printed the incumbent's pid and converged nothing, reopening the hole `convergeInstalledSupervision` exists to close. `Running` and an unreadable status still answer on their own, which keeps a pre-upgrade task (whose action is still the daemon) classified exactly as before.

2. **`daemon install` must end a running task instance before `/Run`.** With `MultipleInstancesPolicy=IgnoreNew`, a `/Run` against an instance of the *previous* action is a silent rc-0 no-op — so on the upgrade the newly registered launcher never ran and the console-window daemon kept serving behind an install that reported success. Under the launcher action there is no instance to end, so this is a migration step exactly once.

Two losses are documented rather than papered over: `RestartOnFailure` now covers only a launcher that cannot spawn (a crashed daemon returns at the next logon or by hand), and `schtasks /End` no longer stops the daemon — `daemon stop` already ended the task *and* POSTed `/shutdown`, and the second one is what did the work.

## The bill for a console-less daemon (second commit)

This PR was closed for a day because the fix had a regression: with the daemon detached, an empty `C:\Windows\system32\taskkill.exe` window started appearing at random on the reporter's desktop. **A console-subsystem child of a console-*less* parent gets a brand new console, and a new console comes with a window.** While the task's action was the daemon, the daemon had a console, children inherited it, and nothing was ever drawn — removing the window is what made the gap visible.

The culprit was the Codex rate-limit probe's five-minute cleanup kill (`codex-rate-limits-live.ts`): `codex.cmd` runs under `shell`, so the child is `cmd.exe` and the real server is its grandchild, hence a `taskkill /T` — spawned without `windowsHide`. Two-arm measurement from a console-less detached parent, 12 `taskkill` spawns each, sampling visible windows twice a second: **138 window sightings without `windowsHide`, 0 with it**.

The opposite fix does not exist: a child spawned with `windowsHide` but *not* `detached` inherits the launcher's console (CREATE_NO_WINDOW is ignored when no new console is created) and dies with it, verified by a probe whose child never wrote its first log line. So the daemon stays console-less and the spawns are what change.

An audit of all 138 `child_process` calls in `bridge/src` found two more real gaps — both `adb reverse` calls in the daemon's poll loop — while everything else either already hid its window or runs a binary with no Windows counterpart (spawn fails ENOENT, draws nothing). **`bridge/src/__tests__/windows-child-window.test.ts` now gates this**: a call must hide its window, name a POSIX-only binary, or carry an inline `windows-hide-exempt:` reason (five do — a POSIX login-shell branch, two terminal-attached build paths, the macOS-only Foundation Models helper). The gate reads only calls reached through a real `child_process` import (two BLE managers define a local `spawnSync`) and blanks comments first, after its own first run reported a `spawn()` out of a prose sentence about EBADARCH.

## What a reboot taught us (third commit)

The morning after, the machine came up with the console window again — and the launcher had not failed, it was **no longer installed**. The task's action had been rewritten at 07:25:48, five minutes before the 07:30:35 logon run, by `agentdeck daemon install` from the *published* 1.3.4 CLI, which has no `daemon autostart` command and can only write the action whose process is the daemon (task-file LastWriteTime and shell history both name it). Worth stating in the docs because it is the ordinary upgrade order on a dev machine: the patched build must be the one the **installed** CLI carries, or any `daemon install` / `npx @agentdeck/setup` silently restores the window at the next logon.

Re-registering from the patched CLI also showed the stamp's one bounded weakness: through the handover `daemon.json` goes missing for a few seconds — the departing daemon deletes the discovery file after its successor has written it — so supervision reads `unsupervised` until `ensureDaemonInfo`'s self-heal rewrites the file. The stamp comes back with it, because it lives in the record the daemon built at startup rather than being computed at read time.

Both findings are recorded in `.claude/rules/daemon-lifecycle.md` and the devlog entry.

## Verification

Local gates: `pnpm build`, `pnpm typecheck`, `pnpm test`, `pnpm devlog:build` + `devlog:check`, `pnpm docs:check`, `pnpm design-system:check` all pass. The test suite has 16 failures on this Windows host that are present on `master` unchanged (darwin-only `pty-manager-spawn-helper` / `python-ble-runtime`, missing Python for `mlx-server-guard` and the protocol generator, and `C:\C:\` `URL.pathname` artifacts in `cli.test.ts`); `design/lint.sh` matches the `master` baseline apart from two hits inside a build artifact. `python3 design/verify-tokens-sync.py` could not run — no Python on this host — and no tokens were touched.

End-to-end on the reporting machine, with the patched build installed as the global CLI:

- install migrated the stale instance (`Ended the previous 'AgentDeckDaemon' instance so the newly registered action can start.`) and converged to `Daemon running under the scheduled task AgentDeckDaemon (PID …, port 9120)`
- `daemon.json` carried `"startedBy": "schtasks"`, `/health` answered `mode: daemon`, task status `Ready`
- a window enumeration found **no** console window for the daemon, only the user's own PowerShell window
- `daemon status`, `daemon stop` (stayed stopped), `daemon start` and `daemon restart` all behaved, the latter two reporting the daemon as supervised
- **after a full reboot: the daemon starts at logon with no console window and no taskbar button**, and no stray child windows since the `windowsHide` fixes

New unit coverage: the task XML runs the launcher and never `--foreground`; `composeSchtasksRunning`'s three readings; `startedBySupervisor` accepting only known supervisor kinds out of the environment; and the `windowsHide` source gate plus unit tests for its four helpers.

### One CI note

The Node 26 job failed on the first run at `hooks/src/__tests__/install.test.ts > delivers UTF-8 stdin through Git Bash and Windows PowerShell`, which hit its `execFile` `timeout: 10_000` on a cold runner (bare "Command failed", empty stderr, file duration 10.45 s). This branch touches no `hooks/` file and the test passes locally in 660 ms; the same flake is recorded in the 1.3.4 devlog, where a rerun passed. Happy to raise that timeout budget here or in a separate PR if you'd like the flake closed.


dudziakl:
Closing while I fix a regression this change surfaces on Windows.

With the autostart daemon now running DETACHED (no console of its own), every console child it spawns without `windowsHide` allocates a brand new console — *with* a window. In practice that means a `taskkill.exe` window popping up on the desktop every Codex rate-limit probe cycle (and the same for the `adb reverse` poll). The window that used to host the daemon was hiding this: children inherited it and drew nothing new.

I confirmed the alternative does not work — a child spawned with `windowsHide` but not `detached` inherits the launcher's console (CREATE_NO_WINDOW is ignored there) and dies with it, so the daemon has to stay console-less and the child spawns are what must be fixed.

Will reopen with those spawns fixed and a gate so new ones cannot regress.

dudziakl:
Reopened. I closed this on the 14th because the fix had a regression on my machine — with the daemon detached, an empty `C:\Windows\system32\taskkill.exe` window started appearing at random. That is fixed, and the branch now carries three commits:

| Commit | What |
|---|---|
| `d7cf00e3` | the launcher action, the `startedBy` stamp, the end-before-run migration |
| `d40cf9d4` | `windowsHide` on the spawns that were drawing windows (Codex rate-limit probe's `taskkill /T`, the two `adb reverse` poll calls) plus a vitest gate so the next one can't land |
| `a4beb205` | the two things a live reboot taught the launcher, written into the rule file and devlog |

The cause of the stray window is worth stating plainly because it is the price of the fix: **a console-subsystem child of a console-*less* parent gets a brand new console, and a new console comes with a window.** While the task's action was the daemon, the daemon had a console, children inherited it, and nothing was ever drawn. Measured from a console-less detached parent, 12 `taskkill` spawns per arm: 138 window sightings without `windowsHide`, 0 with it.

Verified across a full reboot on Windows 11 26200: the daemon starts at logon with no console window and no taskbar button, `daemon.json` carries `"startedBy": "schtasks"`, task status reads `Ready`, and no stray child windows have appeared since.

Background and the measurements behind each rejected alternative are in #339.

dudziakl:
@puritysb I need your help with faling test. It seems that there is some Korean character that is failing this check. Can you take a look?

More details about it from Claude
---
The Korean characters aren't a failure — they're the test's fixture, and they're deliberate. Two different things use them:

const home = mkdtempSync(join(tmpdir(), 'agentdeck hook 한글-'));           // path: space + non-ASCII
const payload = JSON.stringify({ session_id: 'review', prompt: '안녕하세요 café' });  // body: non-ASCII
So one test exercises both hostile cases at once: a hook-script path containing a space and non-ASCII characters (survives Git Bash quoting → powershell -File "…"), and a non-ASCII payload surviving the stdin→HTTP trip.

The bug it guards is [#46](https://github.com/puritysb/AgentDeck/issues/46), fixed in 3217145c (2026-07-10). The Windows Claude hook mangled any non-ASCII payload in two independent places, both PowerShell 5.1 defaults:

[Console]::In.ReadToEnd() decodes piped stdin with the console OEM codepage — CP949 on a Korean-locale machine — so 안녕하세요 was already garbage before it left the script.
Invoke-RestMethod with a string body and no charset in the content type encodes the request as ISO-8859-1, which has no code points for those characters and substitutes ?.
The fix, still in [hooks/src/install.ts:171-176](vscode-webview://1ae6uq2bi3e91qpb3rtf89rfn550rbl0n8ubs18ha8i21tt379c7/hooks/src/install.ts#L171-L176), pins both ends explicitly:


$body  = (New-Object System.IO.StreamReader([Console]::OpenStandardInput(), [System.Text.Encoding]::UTF8)).ReadToEnd()
$bytes = [System.Text.Encoding]::UTF8.GetBytes([string]$body)
# … -Body $bytes -ContentType 'application/json; charset=utf-8'
It was a follow-up: the Codex hook path already did this, and the Claude one-liner had been left behind — which is exactly the failure mode the "mirrored byte-identically in three places" rule in CLAUDE.md exists to prevent.

On the CI failure specifically: the Korean path is not what broke it. Error: Command failed with empty stderr and a 10.36 s file duration against timeout: 10_000 is the timeout firing. And the encoding surface is identical in all three Node lanes — Node 24 and 26 passed the same test in the same run, and it passed on Node 26 in the previous run while failing there. A codepage or path-quoting defect would fail every lane, every time; hopping lanes between runs is what a cold-runner timeout looks like. The test spawns bash.exe → powershell.exe → an HTTP round trip, which is a slow chain to fit in 10 s on a cold Windows runner.

puritysb:
Pushed maintainer follow-up 3bb1666e on top of your three commits.

- Preserve `unknown` if `daemon.json` cannot be read. The handover gap you measured cannot safely be treated as `unsupervised`: install stops an answering daemon immediately on that verdict, before registry self-healing can help. A readable unstamped record still means unsupervised.
- Await the child's `spawn`/`error` event before the launcher exits. An asynchronous creation failure now exits nonzero, allowing the task's launcher retry policy to work.
- Keep the Korean path and UTF-8 payload assertions. The previous 10-second failure suggests a timeout, but its log did not expose the signal/killed fields, so it does not prove the cause. The encoding E2E now allows 30 seconds for cold Git Bash/PowerShell startup and reports elapsed time, code, signal, killed status and whether HTTP arrived on failure. Production hook timeouts are unchanged.
- Run the new launcher and existing supervisor regressions on all three Windows Node lanes.

Local build/typecheck passed; all 293 test files passed (4,525 tests, 3 platform skips). Protocol generation left no tracked drift; docs/devlog/catalog/token checks passed. Design lint remains at its clean-checkout baseline of 89 existing violations. Updated lifecycle documentation and devlog are included.

Waiting for the new Windows CI results. The detached daemon still has the documented limitation that Task Scheduler retries launcher failures, not later daemon crashes. This branch is not published to npm; a repeat Windows logon/install/stop/start/restart check must use the patched build.

Update: merged current master into this branch as 42c7ed98 and regenerated the devlog from both sides' entries to resolve the PR conflict. No executable changes came from that merge; docs/devlog/catalog checks passed again. CI is now running: https://github.com/puritysb/AgentDeck/actions/runs/34954096937.

Final CI update: head 2a09e90f is green in all six checks, including Windows Node 22/24/26 and Linux full CI: https://github.com/puritysb/AgentDeck/actions/runs/34954367636. Expanding Windows coverage exposed one existing systemd test fixture inheriting the runner's Windows data-directory path; 2a09e90f gives that Linux-rendering fixture explicit Linux paths. The UTF-8 hook test and both new launcher tests pass on Windows.

Please repeat your Windows logon/install/stop/start/restart check against this final head when available, especially the no-console behavior after waiting for `spawn`. The PR remains open and no npm version has been published from it.

puritysb:
Merged and delivered in npm 1.3.5. All four package versions and latest tags are publicly verified against the tested release source. Upgrade with `npx @agentdeck/setup@1.3.5 --yes` to re-register Windows autostart using the patched installed CLI. The original contributor commits are preserved. Windows Node 22/24/26 regressions and installed Mac/Lenovo daemon-transition checks passed; Windows Task Scheduler still retries launcher creation failures rather than later detached-daemon crashes, as documented.

dudziakl:
Thank you❤️

PR #346, opened 2026-09-18
feat: show Luna reserve on Codex usage surfaces

puritysb:
## What

Codex accounts with a Luna reserve report it as an additional rate-limit pool the account 5h/7d windows do not describe. This surfaces it end to end on the TS-driven surfaces:

- `shared/src/protocol.ts`: new `CodexLunaReserve` (`lunaReserve` on `CodexRateLimits`); Swift/Kotlin/JSON protocol mirrors regenerated (`pnpm generate-protocol`, no drift).
- `bridge/src/codex-rate-limits-live.ts`: `parseLiveCodexRateLimits` extracts the pool from `additionalRateLimits` (falling back to the legacy `rateLimitsByLimitId` reserve block); `pickBestCodexRateLimits` preserves the live read's Luna metadata when the newer passive rollout wins the window race — rollouts cannot carry additional pools. A plan-mismatched snapshot is still voided whole (Luna included) by `normalizeCodexRateLimits`, so a retired reserve cannot pin a gauge on retain-on-absent clients.
- Plugin: while a reserve is reported it **replaces** the Codex 5h/7d gauges — session buttons (crescent moon tile, remaining % or EMPTY, reset countdown) and the SD+ Codex encoder LCD.
- D200H: shared layout engine renders the same Luna tile in place of the Codex gauges; `plugin-ulanzi` deck signature invalidates on Luna appearance/change/disappearance.
- Apple device preview: `D200HLayoutModel.swift` re-ported (Luna semantics + crescent drawing in `DeskPreviews.swift`), SYNC-HASH pin advanced, snapshot test added.

## Review notes

- Luna matching is heuristic on the upstream shape (`meteredFeature`/`limitName` containing "luna", or a `reserve`/`base_model_inference` block) — the endpoint is not a public versioned contract.
- `keepLuna` grafts without a family gate: both snapshots describe the same account, and plan-mismatch voiding happens later in `normalizeCodexRateLimits` (comment states this explicitly).

## Scope limits (intentional)

Apple, Android and ESP32 surfaces do not render Luna yet — the Swift usage parser does not emit the field and the firmware is unchanged. The wire field is optional, so those clients keep their current Codex gauges.

## Validation

- `pnpm build` + `pnpm typecheck` clean; 4,539 vitest tests pass (2 platform skips), incl. new parser / merge-preservation / tile / encoder / deck-signature cases.
- `pnpm generate-protocol` leaves no drift; preview-mirror SYNC-HASH gate green after the re-port.
- macOS `DevicePreviewSnapshotTests` pass under xcodebuild, incl. new `testLunaReserveReplacesCodexWindowsWithOneTile`.
- Token mirrors in sync. `design/lint.sh` reports 91 pre-existing violations in docs HTML untouched by this branch (non-gating in CI, base-inherited).
- Devlog entry added; `pnpm devlog:check` + `pnpm docs:check` pass.


plugin/src/session-slot-manager.ts
```diff
@@ -494,7 +498,15 @@ export class SessionSlotManager {
...
-    for (const w of codexWindows) {
+    if (this._codexLunaReserve) {
+      gauges.push({
+        agent: 'codex', window: '5h', label: 'LUNA',
+        percent: this._codexLunaReserve.usedPercent,
+        resetsAt: this._codexLunaReserve.regularResetsAt ?? this._codexLunaReserve.resetsAt,
+        known: true, color: CODEX_USAGE_COLOR, luna: this._codexLunaReserve,
+      });
+    }
+    for (const w of this._codexLunaReserve ? [] : codexWindows) {
```
puritysb, line 509: **Test gap: the Luna replacement semantics are untested at the TS origin.**

`this._codexLunaReserve ? [] : codexWindows` here and `cx?.lunaReserve ? [] : codexWindowsBeside(...)` in `shared/src/d200h-layout.ts:516` are the only implementations of "a reported reserve replaces the Codex 5H/7D gauges", but no TS test exercises them — `d200h-layout.test.ts` only asserts the tile SVG in isolation, and neither `session-deck-usage.test.ts` nor `session-slot-manager.test.ts` has a Luna case at all.

The behavior is currently gated only by the Swift mirror (`testLunaReserveReplacesCodexWindowsWithOneTile`), whose doc comment says it "Mirrors `session-deck-usage.test.ts` / `d200h-layout.test.ts` on the TS side" — neither file actually has one. If the TS ternaries are later dropped or broken, no TS test fails while the Swift snapshot keeps passing against its own port.

Could you add a TS case (in `session-deck-usage.test.ts` or `session-slot-manager.test.ts`) asserting: with `lunaReserve` present the Codex 5H/7D gauges are replaced by exactly one LUNA gauge/tile (Claude windows untouched), and without it the windows return?

bridge/src/codex-rate-limits-live.ts
```diff
@@ -151,6 +172,14 @@ export function parseLiveCodexRateLimits(result: unknown, capturedAt: string): C
     planType: typeof rl.planType === 'string' ? rl.planType : undefined,
     limitId,
     credits,
+    lunaReserve: lunaPrimary || lunaSecondary
```
puritysb, line 175: **Minor: the Luna window never gets the stale treatment that account windows get.**

Account windows pass through `normalizeCodexWindow` (`usage-event.ts`), which drops a lapsed `resetsAt` and flags `stale: true` so surfaces dim/drop the gauge — axis (1) of the four-axes rule in `.claude/rules/usage-quota.md`. The reserve's `resetsAt`/`regularResetsAt` bypass that entirely: once the reserve's reset lapses, the tile keeps rendering "N% LEFT / RESET IN now" until the next successful live poll refreshes or drops the block.

Self-correcting and the reserve resets on a slow cadence, so the impact is low — but this is the one window shape that can render past its own reset as if live. Either run the reserve's reset(s) through the same stale handling in `normalizeCodexRateLimits`, or note the omission as intentional here.

bridge/src/codex-rate-limits-live.ts
```diff
@@ -654,15 +687,23 @@ export function pickBestCodexRateLimits(
...
     accountPlan,
   )
     ? live
     : passive;
+  // The passive rollout is normally newer and wins the account-window race,
+  // but it cannot carry additional pools. Preserve Luna metadata from the live
+  // account read — both snapshots describe the same account, so no family
+  // gate is needed here; a plan-mismatched snapshot (its own Luna included)
+  // is voided later by normalizeCodexRateLimits.
+  return keepLuna(chosen, chosen === live ? passive : live);
```
puritysb, line 706: **Minor: the grafted Luna has no age axis of its own.**

When the passive rollout wins the window race, `keepLuna` grafts the live read's Luna onto a snapshot whose `capturedAt` is the passive's — newer than the Luna reading itself. The reserve percent therefore renders as fresh when it is actually as old as the last live read, and under live-fetch failure backoff that skew can grow beyond one poll interval.

The comment covers the family dimension (why no family gate is needed) but not this one. The skew is bounded by the live poll cadence and the reserve drifts slowly, so accepting it is fine — but the reasoning should live next to the graft, e.g. "the grafted percent may trail the frame's `capturedAt` by up to the live poll interval; accepted because …". (Emitting the reserve's own `capturedAt` would be the fuller fix, but that is beyond this PR.)

puritysb:
Addressed in 27cd8b0f — the replacement semantics are now pinned at both TS origins:

- `shared/src/__tests__/session-deck-usage.test.ts` — `replaces the Codex windows with one LUNA tile while a reserve is reported`: with `lunaReserve` present, exactly one LUNA tile seats in place of the Codex pair (Claude 5H/7D untouched, displaced windows' percents render nowhere, exhausted reserve reads EMPTY); the same report without the reserve seats both windows again.
- `plugin/src/__tests__/session-slot-manager.test.ts` — same toggle for the SD keypad: 3 usage keys with one LUNA gauge carrying `usageLuna`/`usagePercent` from the reserve, 4 keys with the Codex 5H/7D pair when it is absent.

Full suite green: 4,541 passed / 2 platform skips; typecheck clean.

puritysb:
The earlier `Windows native runtime (Node 24)` failure was a transient runner-side wedge, not a branch regression: `install.test.ts > delivers UTF-8 stdin through Git Bash and Windows PowerShell` hit its 30 s exec timeout with no request received and empty stdout/stderr — the bash→powershell chain stalled before the hook's stdin read completed. The branch touches none of `hooks/`, the same test is green on the merge-base and on Node 22/26 in the same run, and the rerun of the failed job passes. No code change warranted; noting the signature in case it recurs.

plugin/src/session-slot-manager.ts
puritysb: Verified in 27cd8b0f — both origin files now pin the replacement semantics exactly: with the reserve the displaced window percents render nowhere, the exhausted reserve reads EMPTY, and the same report without it seats both Codex windows again. CI green on the head. Resolving.

